PR #25, opened 2023-05-20
Connecting the client to the keria cloud agent and able to retrieve identifiers.

AlexAndrei98:
TODO: Check signature from keria


rodolfomiranda (approved):
LGTM

PR #41, opened 2023-05-25
Incept salty managed keys and post to KERIA

rodolfomiranda:
Still pending:
- [x] passing of `...kargs`
- [x] improve `identifiers.create` method when arguments are `null`

### DIFF
--- a/examples/signify-react-ts/src/App.tsx
+++ b/examples/signify-react-ts/src/App.tsx
@@ -29,9 +29,6 @@ function App() {
     return (
         <>
             <Signify/>
-            <p className="read-the-docs">
-                Click on the Vite and React logos to learn more
-            </p>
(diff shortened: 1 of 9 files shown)


AlexAndrei98:
Looks like the Manager is breaking the Habery. Do we even need Habery? Otherwise it looks good to me

rodolfomiranda:
> Looks like the Manager is breaking the Habery. Do we even need Habery? Otherwise it looks good to me

It's not using habery at all. Where it breaks?

rodolfomiranda:
I found the error and fixed. Will be available in next commit

rodolfomiranda:
Fixed some errors and added rotation

AlexAndrei98:
tested and it work

PR #51, opened 2023-06-14
Add support for libsodium Tier

rodolfomiranda:


src/keri/core/salter.ts
```diff
@@ -4,7 +4,7 @@ import { Matter, MtrDex } from './matter';
 import { EmptyMaterialError } from './kering'
 import libsodium  from 'libsodium-wrappers-sumo';
 
-export const enum Tier {
```
AlexAndrei98, line 7: I made it a class in the [CredentialType](https://github.com/WebOfTrust/signify-ts/blob/main/src/keri/app/signify.ts#L13) example as I was looking at the Codex class

PR #57, opened 2023-07-10
Node support for library and example Node script.

pfeairheller:
Update to index.ts to load Blake3 WASM module and redefine Buffer in window only if in browser.  This enables the unit tests to pass again and also this code to run in Node.

Added a scripts directory under examples which allows has a client script that creates an Agent and an AID in that agent using this library with the following command:

npx --package=signify-ts ts-node client.ts

### DIFF
--- a/examples/scripts/client.ts
+++ b/examples/scripts/client.ts
@@ -1,0 +1,53 @@
+let signify:any;
+
+// @ts-ignore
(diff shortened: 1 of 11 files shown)


rodolfomiranda (approved):
Looks great to me

PR #83, opened 2023-08-10
Update README.md

daidoji:
Fixed typo from "ed25510" to "ed25519".


m00sey (approved):

m00sey:
Thanks! Do you think we can coin the term ed25510? 😬

PR #84, opened 2023-08-15
Fix broken union of Sets

rodolfomiranda:
Fixing a formula used to create a set by the union of two sets that worked in some cases, but not in others.


AlexAndrei98 (approved):
LGTM

PR #104, opened 2023-09-08
Fix integration scripts type checking

psteniusubi:
I think there's an anti-pattern in the integration tests in singify-ts/examples 

All of the .ts files start with this declaration
`let signify: any;`

This effectively disables any type checking by the typescript compiler, VS Code and other tools

This pull request is a suggestion how to fix this. If this pattern is acceptable then the other scripts should be fixed as well.

### DIFF
--- a/examples/scripts/client.ts
+++ b/examples/scripts/client.ts
@@ -1,23 +1,12 @@
-// @ts-ignore
-let signify: any;
+import signify from "signify-ts";
 
-// @ts-ignore
-import('signify-ts').then(
-    (module) => {
-        signify = module
-        signify.ready().then(() => {
-            console.log("Signify client ready!");
-            connect().then(() => {
-                console.log("Done")
-            });
-        });
-    }
-)
+await connect();
 
 async function connect() {
     let url = "http://127.0.0.1:3901"
     let bran = '0123456789abcdefghijk'
 
+    await signify.ready();
     const client = new signify.SignifyClient(url, bran);
     console.log(client.controller.pre)
     const [evt, sign] = client.controller?.event ?? [];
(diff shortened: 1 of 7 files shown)


pfeairheller (changes requested):
Yes, this looks like a good approach.  Please add the fix to all the other scripts and we can get this merged.

psteniusubi:
@pfeairheller 

If fixed this for make_endroles.ts and list_notifications.ts. However in the latter there is now a compile error that I cannot easily find a solution for

```
list_notifications.ts:74:45 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'Uint8Array'.

74                     let esigs = keeper.sign(nexn.raw)
                                               ~~~~~~~~
```

pfeairheller:
> @pfeairheller
> 
> If fixed this for make_endroles.ts and list_notifications.ts. However in the latter there is now a compile error that I cannot easily find a solution for
> 
> ```
> list_notifications.ts:74:45 - error TS2345: Argument of type 'string' is not assignable to parameter of type 'Uint8Array'.
> 
> 74                     let esigs = keeper.sign(nexn.raw)
>                                                ~~~~~~~~
> ```

There is a method in Signify called `b` that will convert a string to a Uint8Array, use that on `nextn.raw

psteniusubi:
Thanks @pfeairheller that was the solution. I also had to change require("prompt-sync") into import. 
I tested and I can complete the multi-sig group inception sequence.
I'll create an other pull request for the scripts in examples/integration-scripts

rodolfomiranda:
I used the same approach in all scripts under `/examples/integration-scrpts`. PR #105

pfeairheller (approved):

PR #107, opened 2023-09-12
Reafactor signify ts

AlexAndrei98:


src/keri/app/clienting.ts
```diff
@@ -0,0 +1,454 @@
...
+    /**
+    * Get exchange resource
+    * @returns {Exchanges}
+    */
+    exchanges(): Exchanges {
+        return new Exchanges(this)
+    }
+}
+
+export class EventResult {
```
rodolfomiranda, line 431: I think this should go in aiding.ts since `EventResult` represent the result of an `icp`, `ixn` or `rot` event.

rodolfomiranda (changes requested):

rodolfomiranda (approved):

PR #108, opened 2023-09-22
use Response.ok to check http response status

psteniusubi:
Was not accepting `204 No Content` status for example


rodolfomiranda (approved):
that's the way

PR #109, opened 2023-09-23
Add support for SHIM external modules

rodolfomiranda:


examples/integration-scripts/bip39_shim/src/bip39_shim.ts
```diff
@@ -0,0 +1,113 @@
...
+        this.icount = this.ncount
+        this.ncount = ncount
+
+        let nsigners = this.keys(this.ncount, this.kidx + this.icount, this.transferable)          
+        let digers = nsigners.map(nsigner => new Diger({code: this.dcode},nsigner.verfer.qb64b ).qb64);
+
+        return [verfers, digers]
+    }
+
+    sign(ser: Uint8Array, indexed=true, indices:number[]|undefined=undefined, ondices:number[]|undefined=undefined){
```
AlexAndrei98, line 71: is this the function that gets called when you need to sign icp,rot,exn(etc) events etc?

AlexAndrei98 (approved):

examples/integration-scripts/bip39_shim/package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "bip39_shim",
+    "type": "module",
+    "private": true,
+    "version": "0.0.1",
+    "scripts": {},
+    "dependencies": {
+        "bip39": "^3.1.0",
```
AlexAndrei98, line 8: I think we should put the shim in a different place and import it here.

examples/integration-scripts/bip39_shim/src/bip39_shim.ts
rodolfomiranda: yes, this is a key manager like the `SaltyKeeper` or `RandyKeeper`. They need to implement the `incept`, `rotate`, `sign` and `param`.

rodolfomiranda:
Be aware that this PR changes some functions to become `async` like `identifiers.create` and `identifiers.rotate`. The scripts were adapted with the `await` operator and were validated to execute successfully.

m00sey:
@rodolfomiranda can you resolve the conflicts here?

PR #115, opened 2023-10-05
(draft) Ipex part deux

kentbull:


pfeairheller:
Can this please be resubmitted without all the reformatting, please.

It is hard to tell what substantial changes are in the PR.

examples/integration-scripts/multisig-credentials.ts
m00sey, line 1: My expectation was that credential create... etc would follow on in multisig.ts where registry create ended.

Yes, the file is big but its a single readable flow, not sure we need another flow for it.

PR #118, opened 2023-10-16
Credential issuance and IPEX grant message for multisig participants

pfeairheller:
This PR includes:

* Implementation of IPEX grant message
* Updates to credential issue to support multisig issuance
* A new script located in examples/scripts called multisig-create-credential.ts that is a peer to the exact same script (but ends in .py) in SigPy.  It can be run with multisig-signify-issue.sh from KERIpy to create a multisig issuer, create a credential registry, create a credential and GRANT it to the holder via IPEX as a participant in a 3 way multisig AID.


m00sey (approved):
🚀

m00sey:
Happy to merge this and I can work on coverage

PR #119, opened 2023-10-17
remove conflicting prettier config in package.json

lenkan:
There was already a conflicting prettier config in package.json. Some editors might use this instead of the .prettierrc file which causes formatting conflicts when using "format on save" vs "npm run pretty".


package.json
m00sey, line 1: what was the actual change here?

Hard to see between the reformatting

lenkan: Sorry, it was the indentation. My editor applied the .prettierrc indentation of 4 spaces instead of 2 spaces. I fixed it by running `npx prettier --tab-width 2 package.json package-lock.json`.

m00sey (approved):

PR #126, opened 2023-11-02
add script for single issuer and single holder ipex

lenkan:
Adding reference script for issuing a credential using a single issuer and a single holder. Includes updates to `ipex().grant()` from @kentbull.

Original post:

> Referring to the discussion from discord today. Here is a script that performs credential issuance with IPEX grant + admit with a single issuer and a single holder.
> 
> This script stops at line L266 printing "No credentials yet..." because the holder never gets any credentials from its agent.
> 
> Do you have any pointers to what I am doing wrong here? I believe I am missing a crucial step in either the "issueCredential" or "admitCredential" function.
> 
>


lenkan:
Thanks @pfeairheller for the updates and adding the /ipex/admit to KERIA. I have tried this again by porting the "submitAdmit" function to here.

- https://github.com/WebOfTrust/signifypy/pull/93
- https://github.com/WebOfTrust/keria/commit/f80d9766f2041667d20510dc8fa26aed0c9fe9e4

However, KERIA now becomes unresponsive after posting the /ipex/admit message. After debugging I found that execution stops in the Admitter when calling "parseOne" on the "iss" embed: https://github.com/WebOfTrust/keria/commit/f80d9766f2041667d20510dc8fa26aed0c9fe9e4#diff-a169faf61f8a8fdf3be4d9d2a164f646003d0f5203efb18f599f4b59393327efR527

`parser.parseOne` simply never returns. It gets stuck in the infinite loop here: https://github.com/WebOfTrust/keripy/blob/6ab521ec83f1bfd67882199fea4c5f45d91304c4/src/keri/core/parsing.py#L418

rodolfomiranda (approved):

Issue #127, opened 2023-11-03
Publish npm package on push


lenkan:
It would be nice to have the signify-ts package published to npm (or github packages) on push. We can use a "development" or "next" tag for pushes to development branch. 

The benefit of this is that users avoid pulling all signify-ts build dependencies and just use the already built modules. I.e., only the dist folder and package.json files will be published.

I  can create an outline PR with the workflow for this if you think it is a good idea.

kentbull:
I will be setting this up so we can have automated releases so we don't have to work so hard to get new versions published.

PR #128, opened 2023-11-03
remove unused dependencies, move dev dependencies

lenkan:
Clean up dependencies as a preparation for a future npm publish. (#127)

- Moved testing and build dependencies from "dependencies" to "devDependencies" to avoid having development dependencies shipped with published package. Also allows us to run `npm audit --omit=dev` to run audit on only runtime dependencies if we want.

- Remove dependencies not used from root package.json

- Remove typescript and eslint dependencies from the examples/scripts and examples/integration-scripts packages. They will just use the typescript and ts-node packages from the root package using nodes module resolution. I think this is better to avoid having multiple versions of typescript installed here.

- Upgrade typescript to 5.1. Latest is 5.1. The scripts used tsconfig settings that were not intruduced until v5, but tsdx or puppeteer had dependencies requiring <= 5.1.


rodolfomiranda:
great clean up. Thanks!

rodolfomiranda (approved):
LGTM

PR #124, opened 2023-10-30
Add credential and ipex to multisig integration script

rodolfomiranda:

PR #133, opened 2023-11-11
Multisig grant and admit messages and fixes in seqner

rodolfomiranda:
This PR replaces PR #124. It adds examples on the integration script `multisig.ts` showing how to pass `grant` and `admit` messages between a multisig issuer and a single sig holder.
It also fixed a serialization error on the seqner and holder that used little endian instead of big endian. Test results were validated against tests in `keripy`

### DIFF
--- a/examples/integration-scripts/multisig.ts
+++ b/examples/integration-scripts/multisig.ts
@@ -427,9 +427,7 @@ async function run() {
 
     // Check for completion
     op1 = await waitForOp(client1, op1);
-
     op2 = await waitForOp(client2, op2);
-
     op3 = await waitForOp(client3, op3);
     console.log(`End role authorization for agent ${eid1}completed!`);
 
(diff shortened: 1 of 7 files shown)


examples/integration-scripts/multisig.ts
```diff
@@ -197,7 +105,7 @@ async function run() {
     // First member start the creation of a multisig identifier
     let rstates = [aid1['state'], aid2['state'], aid3['state']];
```
nkongsuwan, line 106: When three clients are hosted on different machines, or even on different shells, they will not be able to directly obtain other clients' states directly like in this examples.

I suggest adding `client.keyStates().get(...)` for each client to obtain other clients' states.

rodolfomiranda: You are correct, in a real scenario you should keep track of the states independently.

pfeairheller (approved):

PR #134, opened 2023-11-11
use esbuild, tsc and eslint instead of tsdx

### LINKED ISSUE #130: Consider moving away from tsdx
tsdx seems to be largely an abandoned project. See

- https://github.com/jaredpalmer/tsdx/issues/1058
- https://github.com/jaredpalmer/tsdx/issues/1065

It will probably cause some headaches going forward upgrading node versions, testing frameworks, linting and so on. I suggest removing it and just use `tsc` directly.

We probably do not need to build minified builds in the npm package. `signify-ts` can be published as an ES Module package. Then it is up to the user of the package to decide if they want to bundle it with their application code, for example if they are building a web app.

lenkan:
Resolves #130

This PR removes `tsdx` in favor of using `tsc`, `eslint`, `jest` directly. 

## Bundling

`tsdx` used `rollup` for bundling. I opted for `esbuild` which is a more modern bundler and is used by `aws-cdk` and probably some other big stable projects. Ideally, I would like to avoid bundling altogether and just use `tsc` to transpile the typescript sources to runnable javascript. However, the dynamic import of `blake3` in the `ready` function caused some issues with this. We can revisit at a later time to simplify this.

## npm workspaces

Using npm workspaces to simplify dependency management of the examples/ directory. Read more here https://docs.npmjs.com/cli/v10/using-npm/workspaces. Essentially, now you only need to do `npm install` in the root directory to get all the dependencies needed to run the scripts and frontend app.

Note that when you install the project from github (.e.g., `npm install WebOfTrust/signify-ts#development`)  you still only get the root dependencies, so no unnecessary dependencies will be installed.

## Remove `text-encoding` package

This repository has been archived since 2018: https://github.com/inexorabletash/text-encoding and caused issues when bundling because of the way the package was imported. This package should not be needed though. The TextEncoder API is supported by all modern browsers and node.js. See: https://caniuse.com/textencoder.

## Restructured package export files

Previously, rollup did some magic to bundle all named exports into a default export. This is not supported by most other bundlers as far as I am aware. So I did it manually by gathering all named exports to `export.ts`, then re-exporting in `index.ts` while also grouping them as the `default` export. This ensures no existing code will break. However, we should revisit this as well and reconsider what is exported from the library.

## Linting

I added the "recommended" `.eslintrc` config and just lowered all reported errors to "warning" for now. We should create an issue to resolve the errors. I just did not want this PR to blow up.


kentbull:
Nice work. I wanted to upgrade to the latest TextEncoder as well. Glad to see this moving forward.

rodolfomiranda (approved):

PR #136, opened 2023-11-14
add reproduction script for keria crash

lenkan:
This is just a script to reproduce the keria crash https://github.com/WebOfTrust/keria/issues/125. It is probably not a good script for demo purposes but could be used as an integration test when fixed to ensure that keria can properly handle this error without crashing.


lenkan:
I think we can just close this then.

PR #137, opened 2023-11-14
Update an integration example for credentials to support ipex

nkongsuwan:
I merged Kent's [commit](https://github.com/WebOfTrust/signify-ts/commit/cd401c2cd4676e31a79d3c339698e76f8e647a30) and renamed the branch.

This updates the integration example for issuing and presenting credentials with IPEX. Currently, revocation has not yet beencompleted.

### DIFF
--- a/examples/integration-scripts/credentials.ts
+++ b/examples/integration-scripts/credentials.ts
@@ -247,23 +247,23 @@ async function run() {
     console.log('Grant message sent');
 
     // Recipient check issued credential
-    let grantNotification = null;
-    while (grantNotification == null) {
+    let grantNotification1 = null;
+    while (grantNotification1 == null) {
         let notifications = await client2.notifications().list();
         for (let notif of notifications.notes) {
             if (notif.a.r == '/exn/ipex/grant') {
-                grantNotification = notif;
+                grantNotification1 = notif;
             }
         }
         await new Promise((resolve) => setTimeout(resolve, 1000));
     }
 
     dt = new Date().toISOString().replace('Z', '000+00:00');
-    const [admit, sigs, aend] = await client2.ipex().admit('recipient', '', grantNotification.a.d!, dt);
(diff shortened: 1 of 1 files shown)


examples/integration-scripts/credentials.ts
```diff
@@ -3,287 +3,428 @@ import signify from 'signify-ts';
...
-    assert.equal(creds3[0].status.s, '0'); // 0 = issued
+    const [grant2, gsigs2, gend2] = await holderClient
+        .ipex()
+        .grant(
+            holderAidName,
+            verifierAID,
+            '',
+            acdc,
+            issResult.acdcSaider,
+            iss,
```
kentbull, line 347: This test does not represent what will occur in production due to separation of state for issuer, holder, and verifier.

The problem with the way this test does presentation is that it uses state for the Holder's presentation that comes from the Issuer's original IPEX GRANT. This state includes the Issuer's original signature and anchor to it's KEL. These two pieces of issuer state will not be available to the holder from the holder's wallet in production. 

So the question becomes how do we read this state from the holder's database, if there? Are these two pieces of state even necessary or should the holder be putting it's own signature on the original credential and anchor?

The original anchor and signature attachments are parsed by the holder upon receipt of the credential. If these attachments are stored in the holder's LMDB database then we could pull them out to reconstitute the original attachments.

kentbull (approved):
LGTM

examples/integration-scripts/credentials.ts
kentbull: It's the original attachments we need, not the key states. The original attachments include the issuer's anchor of the credential issuance to their KEL as well as the issuer's signature on the ACDC. Those appear to be available by doing a GET for a credential and setting the response content type to "application+cesr". I haven't tried it yet, though that's what I plan to try.

rodolfomiranda (approved):

PR #143, opened 2023-11-21
running integration tests

lenkan:
Here is a rough draft for creating and running integration tests in github actions.

~~The proposal is to move examples/integration-scripts one by one into the `test-integration` folder. We could just keep them as simple node.js script. But if we run them in a test runner we can get coverage reports, test results reporting and also group tests together more efficiently.~~

~~Currently they depend on me pushing images to docker hub. But if we just start to push images from keripy, keria and vLEI on every commit we can just use the images from there. Another option (very slow) is to pull the source or python packages for each dependency and build them here before running the tests. But then we probably have to add some sort of image caching to make it run reasonably fast.~~

The proposal is to rename the examples/integration-scripts with `.test.ts` suffix one-by-one. That way they will be picked up by the test runner.

Currently, the images are being pulled from my docker hub registry. This is because the images in WebOfTrust were not up to date. Once they are, we can simply swap out for example `lenkan/keria` => `WebOfTrust/keria` in docker-compose.yaml.


test-integration/single-issuer-holder.test.ts
pfeairheller, line 1: The idea was to consolidate to one integration directory, why create yet another?

images/keria/entrypoint.sh
pfeairheller, line 1: We should not create KERIA docker files, etc here in this repo

pfeairheller (changes requested):

images/keria/entrypoint.sh
lenkan: It is using the docker file from keria. This is just an entry point script to be able to set configuration.

lenkan:
@pfeairheller There are no additional images built for this. They are built from keria, keripy and vLEI respectively, without any modifications. See docker-compose.yaml where images are specified. They are just hosted in my docker hub registry because the WebOfTrust registry was not updated. The idea is that it would be swapped for `WebOfTrust/keria`.

The reason I added an entrypoint.sh script, and _mounted_ it in the _container_, is to be able to set configuration on startup. There should be another way to do this. This does not change the image, just the container at runtime. I will comment on the specific line.

test-integration/single-issuer-holder.test.ts
lenkan: Well, it was more to illustrate a way to run the scripts as tests rather than standalone scripts. I do not think it would be a lot of work to move them all over. Maybe a middle ground is to just suffix the file with `*.test.ts` and that would be sufficient to convert it into an integration test. Then I can keep them all in the same directory.

images/keria/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+config_dir="/keria/config"
+config_file="$config_dir/keri/cf/agent.json"
+
+mkdir -p "$config_dir/keri/cf"
+
+cat > "$config_file" <<EOF
+{
+  "keria": {
+    "dt": "$timestamp",
+    "curls": ["$keri_url"]
```
lenkan, line 14: @pfeairheller This line here is why I added an entrypoint script. I needed a way to set the hostname for the keri agent config.

Inside the docker network, the hostname for keria is `keria` (specified by the docker compose service). So when I call `generateOobi` for example, I need keria to generate an oobi with the hostname it has within the docker network.

It would be great if you had another way to do this. It feels a little bit hacky. But I was just trying to be helpful and create something that works.

test-integration/single-issuer-holder.test.ts
lenkan: Updated it now to use the old directoy and just add a .test suffix to indicate it is a an integration test. That will make the test runner pick it up.

images/keria/entrypoint.sh
lenkan: I changed it to mount a `keria.json` config file instead of generating the config file in the entrypoint script. It might be less confusing that way. :+1: Thanks for checking.

pfeairheller:
I created and published the following docker images:

WebOfTrust/keri-witness-demo:1.1.0
WebOfTrust/keri:1.1.0
WebOfTrust/keria:0.1.0 (and latest)
gleif/vlei:0.2.0 (and latest)

@lenkan can you please update your compose file and see if they all work for integration?  Thanks!

lenkan:
> I created and published the following docker images:
> 
> WebOfTrust/keri-witness-demo:1.1.0 WebOfTrust/keri:1.1.0 WebOfTrust/keria:0.1.0 (and latest) gleif/vlei:0.2.0 (and latest)
> 
> @lenkan can you please update your compose file and see if they all work for integration? Thanks!

Updated. It works :+1:. I just used :latest for vlei and keria. Not sure what you think is the best strategy there. It is nice that it always pulls the latest stuff to ensure it works. But it can also cause a bit of a nuisance if keria upgrades and breaks integration tests for unrelated PRs at the client repo.

pfeairheller:
> > I created and published the following docker images:
> > WebOfTrust/keri-witness-demo:1.1.0 WebOfTrust/keri:1.1.0 WebOfTrust/keria:0.1.0 (and latest) gleif/vlei:0.2.0 (and latest)
> > @lenkan can you please update your compose file and see if they all work for integration? Thanks!
> 
> Updated. It works 👍. I just used :latest for vlei and keria. Not sure what you think is the best strategy there. It is nice that it always pulls the latest stuff to ensure it works. But it can also cause a bit of a nuisance if keria upgrades and breaks integration tests for unrelated PRs at the client repo.

I think getting the dev containers pushed on a regular basis will help us address this.  So we can have dev builds use dev containers across the board and published versions rely on matching published versions.

pfeairheller (approved):
Looks good to me!

rodolfomiranda:
@lenkan , are we ready to merge this one, right?

lenkan:
> @lenkan , are we ready to merge this one, right?

@rodolfomiranda It should be good to merge. I can help moving some of the remaining scripts to tests as well after this is merged. When KERIA images are being pushed on builds from KERIA repository with :latest tag, they will be picked up here.

Issue #145, opened 2023-11-22
Setting up delegated single-sig identifier with witness config 


psteniusubi:
The integration test to reproduce this issue is in PR #157 and here

https://github.com/psteniusubi/signify-ts/blob/feat-integration-test-singlesig-dip/examples/integration-scripts/singlesig-dip.test.ts

Setting up a delegate single-sig identifier with witness config never completes.

psteniusubi:
https://github.com/WebOfTrust/signify-ts/pull/157
This PR has an integration test to reproduce this issue

psteniusubi:
This issue was fixed by latest changes to keripy, keria and signify-ts

PR #148, opened 2023-11-29
Example for delegation with multisig 

nkongsuwan:
I created an example for delegation with multisig with one delegator and two delegatees.

However, the last step currently does not work. The delegatees do not receive the approval from the delegator. I suspect it has something to do with the payload in the exchange message between delegatee members.

Edit: The problem is related to [Issue #145](https://github.com/WebOfTrust/signify-ts/issues/145). The script just works if I remove the delegatees' witnesses.

### DIFF
--- a/examples/integration-scripts/delegation-multisig.test.ts
+++ b/examples/integration-scripts/delegation-multisig.test.ts
@@ -1,0 +1,227 @@
+import { strict as assert } from 'assert';
+import signify from 'signify-ts';
+import { resolveEnvironment } from './utils/resolve-env';
+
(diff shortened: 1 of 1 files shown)


psteniusubi:
Hi @nkongsuwan 
This is probably related https://github.com/WebOfTrust/signify-ts/issues/145
Does the script work if you remove the witness config from the delegate?

nkongsuwan:
@psteniusubi Yes, I just removed the witnesses, and the scripts now work!

lenkan (approved):

PR #153, opened 2023-12-01
add npm ci, linting, format check to workflow

### LINKED ISSUE #129: Use npm clean install to respect lock files in ci workflow
Currently, `npm install` is used in the github actions workflow. This does not respect the lock file, which means new versions of packages will be pulled on install.

Instead, `npm ci` should be used: https://docs.npmjs.com/cli/v10/commands/npm-ci

I can create PR for this if you want.

lenkan:
Closes #129 

Also follow up on #122 to ensure build fails if formatting is not done by prettier.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -27,13 +27,23 @@ jobs:
           node-version: '18.12.1'
           cache: 'npm'
 
-      - name: build
+      - name: Install dependencies
         run: |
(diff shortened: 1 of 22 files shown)


pfeairheller (approved):

pfeairheller:
@lenkan - This now has conflict.  Can you please rebase and I'll get it merged?

PR #156, opened 2023-12-01
Fix serializeACDCAttachment to work with the correct anchor per type.

pfeairheller:
This PR uses the correct events for creating the anchors in the embed of grant messages.

### DIFF
--- a/src/keri/app/credentialing.ts
+++ b/src/keri/app/credentialing.ts
@@ -712,7 +712,7 @@ export class Ipex {
         const atc = ims.substring(args.anc.size);
 
         const embeds: Record<string, [Serder, string]> = {
-            acdc: [args.acdc, d(serializeACDCAttachment(args.acdc))],
+            acdc: [args.acdc, d(serializeACDCAttachment(args.iss))],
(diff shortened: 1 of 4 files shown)


src/keri/app/credentialing.ts
```diff
@@ -712,7 +712,7 @@ export class Ipex {
         const atc = ims.substring(args.anc.size);
 
         const embeds: Record<string, [Serder, string]> = {
-            acdc: [args.acdc, d(serializeACDCAttachment(args.acdc))],
+            acdc: [args.acdc, d(serializeACDCAttachment(args.iss))],
```
lenkan, line 715: I am curious what the effects are of this change since it seemed to work before as well?

pfeairheller: So the attached seal on the ACDC needs to point to the issuance event and the attached seal on the issuance event needs to point to the anchoring KEL event.

lenkan (approved):

kentbull:
This makes sense. I'm glad to learn how do do this correclty.

PR #166, opened 2023-12-06
Add credential presentation from held credential to verifier

pfeairheller:
Addition of submitGrant that calls the correct endpoint to perform a credential grant.

Sample in `single-issuer-holder.test.ts` of using data from credentials/list to create a GRANT for presenting a held credential.

This fixes the grant methods to allow for granting from a held credential.  I believe it supercedes #158

### DIFF
--- a/examples/integration-scripts/single-issuer-holder.test.ts
+++ b/examples/integration-scripts/single-issuer-holder.test.ts
@@ -2,7 +2,7 @@ import assert from 'node:assert';
 import signify, {
     SignifyClient,
     IssueCredentialArgs,
-    Operation,
+    Operation, Serder,
 } from 'signify-ts';
 import { resolveEnvironment } from './utils/resolve-env';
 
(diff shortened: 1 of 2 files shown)


lenkan (approved):
Great! From the logs it looks like the failing integration test just needs a new keria image?

pfeairheller:
> Great! From the logs it looks like the failing integration test just needs a new keria image?

Yup, building / publishing them now.

pfeairheller:
Looks like I also need some unit test coverage.  I'm not happy with the arguments I added and will probably collapse them by using an object for each message being passed in.  So instead of `acdc` and `acdcAttachment`, I'm going to change it to `acdc` that takes `sad` and `atc`.

pfeairheller:
Getting stuck in a mess.  Some of the integration tests weren't fixed to run correctly with the "local" environment.  Fixing that to see what else is going on with them.

PR #167, opened 2023-12-07
use waitOperation utility function in multisig test to fail faster

lenkan:
The utility function `waitOperation` has a timeout, this ensure the test fails faster if an operation does not resolve.

### DIFF
--- a/examples/integration-scripts/multisig.test.ts
+++ b/examples/integration-scripts/multisig.test.ts
@@ -5,6 +5,7 @@ import signify, {
     IssueCredentialResult,
 } from 'signify-ts';
 import { resolveEnvironment } from './utils/resolve-env';
+import { waitOperation } from './utils/test-util';
 
 const { url, bootUrl, vleiServerUrl } = resolveEnvironment();
 const WITNESS_AIDS = [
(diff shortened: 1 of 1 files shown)


examples/integration-scripts/multisig.test.ts
```diff
@@ -935,22 +936,22 @@ test('multisig', async function run() {
...
+    op1 = await waitOperation(client1, op1);
     op2 = await client2.keyStates().query(m.prefix, 4);
-    op2 = await waitForOp(client2, op2);
+    op2 = await waitOperation(client2, op2);
     op3 = await client3.keyStates().query(m.prefix, 4);
-    op3 = await waitForOp(client3, op3);
-    op4 = await client4.keyStates().query(m.prefix, 4);
-    op4 = await waitForOp(client4, op4);
+    op3 = await waitOperation(client3, op3);
+    op4 = await await client4.keyStates().query(m.prefix, 4);
```
pfeairheller, line 953: Looks like we have an extra `await` added here.  I guess it still works, but this should be removed.

pfeairheller (changes requested):

pfeairheller (approved):

PR #168, opened 2023-12-07
suggestion: bail on first failing integration test

lenkan:
This is just a suggestion. I think it makes sense to bail on the test run when we encounter the first failing test. This will give you quicker feedback on the build if a test is failing.

Note that this is only for the integration tests.


psteniusubi:
I agree with this. 
In addition I think we should use for example `jest --testTimeout 30000` to specify a fairly long test timeout by default. Then make sure there are no integration tests that "wait forever" for operations or notifications. This because jest reporting of tests failing because of test timeout is quite bad.

Issue #165, opened 2023-12-06
Decide what to do with the examples/signify-react-ts project


lenkan:
By inspection, I can tell the code in here has not been kept up to date with the recent changes, so it will not work in all scenarios. The `Credentials.tsx` file for example is not updated.

Should we keep it? If yes, we probably need to add some steps to the workflow to test that it works. If we do that, we can decide what level to keep it at:

- Static type-checking. Could be sufficient, but currently there are a lot of `any` types which will effectively turn off typechecking.
- Unit testing. Personally I do not see much benefit in this since it is a testing project in itself.
- E2E testing. Could be run with e.g. puppeteer to ensure that the code works in a browser environment. But is probably a bit more work than it is worth at the moment.

kentbull:
I agree that we should keep this example in the codebase. I would be willing to help maintain it. So many recent changes are part of why we haven't updated it. With Phil's recent work https://github.com/WebOfTrust/signify-ts/pull/166 being complete then we can have an end-to-end flow with issue, admit, and present so now is a good time to update the webapp.

Having a sample one in this repo seems like a good idea and I agree that the tests for this sample app should be run along with tests for the repo.

pfeairheller:
Discussed on 12/7 dev call, decision to move to its own repo.

kentbull:
This is addressed by https://github.com/WebOfTrust/signify-ts/pull/170

PR #170, opened 2023-12-07
Move example app to other repo

kentbull:
The example React and Typescript application has been moved to the WebOfTrust/signify-react-ts repository


kentbull:
Fixes https://github.com/WebOfTrust/signify-ts/issues/165

lenkan:
@kentbull Can you also do an `npm install` in the root directory to refresh the `package-lock.json`-file? Since this is now an npm workspace, the dependencies will then be removed from there.

kentbull:
@lenkan yes, I'll do that as a part of my next merge request

PR #172, opened 2023-12-07
update package-lock after removal of signify-react-ts

lenkan:
Since this repository is an npm workspace, the package-lock file needs to be updated as mentioned in #170 . This is done by running `npm install`.


m00sey (approved):
thanks for catching that

PR #176, opened 2023-12-09
remove endless loops from tests

### LINKED ISSUE #171: Remove endless loops in the integration tests
Currently, some integration tests will enter endless loops to wait for certain operations or assertions. We should make sure that all these loops have timeouts. Otherwise tests may run forever and cause confusing failures. See this PR for an example https://github.com/WebOfTrust/signify-ts/pull/167.

lenkan:
Improves wait/retry function with exponential back-off for faster test run. Looks like the total test run time decreases with about ~20%.

Closes #171

### DIFF
--- a/examples/integration-scripts/delegation-multisig.test.ts
+++ b/examples/integration-scripts/delegation-multisig.test.ts
@@ -1,6 +1,7 @@
 import { strict as assert } from 'assert';
 import signify from 'signify-ts';
 import { resolveEnvironment } from './utils/resolve-env';
+import { waitForNotifications, waitOperation } from './utils/test-util';
 
 const { url, bootUrl } = resolveEnvironment();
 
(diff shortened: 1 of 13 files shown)


examples/integration-scripts/utils/retry.ts
```diff
@@ -0,0 +1,51 @@
+import { setTimeout } from 'timers/promises';
+
+export interface RetryOptions {
+    maxSleep?: number;
+    minSleep?: number;
+    maxRetries?: number;
+    timeout?: number;
+}
+
+export async function retry<T>(
```
kentbull, line 10: I like this abstraction.

psteniusubi:
Discussion about the `retry` function
There may be scenarios where the inner `fn` function wants to signal to the retry polling loop that the operation will never be successful. Currently I don't think this is possible.
Suggestion:
- change fn signature to `fn: () => Promise<T | undefined>`
- change retry polling loop to cancel and re-throw any exception from `fn`
- continue polling loop only when return value is `undefined`

lenkan:
> Discussion about the `retry` function There may be scenarios where the inner `fn` function wants to signal to the retry polling loop that the operation will never be successful. Currently I don't think this is possible. Suggestion:
> 
> * change fn signature to `fn: () => Promise<T | undefined>`
> * change retry polling loop to cancel and re-throw any exception from `fn`
> * continue polling loop only when return value is `undefined`

Thanks for the feedback. Can you elaborate with a use case for when you would want to do that?

As for the API, I think something like this would be less ambiguous than returning undefined:


```
interface RetryContext {
   /**
     * Signals that we want to abort the retry
     */
   abort(): void;
}

async function retry(fn: (context: RetryContext) => Promise<T>): Promise<T> {
    ...
}
```

And use it like this:

```
// ...in a test

await retry(async ctx => {
   ...
   if(someConditionThatIndicatesStopRetrying) {
      ctx.abort()
   }
})
```

Or another alternative is to use an Error subclass and use it like this:

```
// ...in a test

await retry(async () => {
   ...
   if(someConditionThatIndicatesStopRetrying) {
      throw new AbortRetryError("Some reason")
   }
})
```

But still, I think it would be good to have a use case for it. Maybe it can simply be solved by having a shorter default timeout.

psteniusubi:
For the first alternative, allow passing an exception: `ctx.abort(new Error("critical error"))`

[AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) and AbortSignal could also be used in the first alternative. 

As for the use case. I don't have anything ready. For example there exist `DELETE /operation/{name}` I'm using this to remove and clean up completed operations (part of my waitOperation). I have run into race condition scenarios cases where `GET /operation/{name}` returns HTTP error 404. Logical error of course, but nice to catch early.

lenkan:
> [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController) and AbortSignal could also be used in the first alternative.

Yes, perhaps that is the most recognizable API since `timers/promises` and `fetch` and other APIs already support AbortSignal. Then it would be:

```
const controller = new AbortController();

... in test
await retry(async () => {
   ... Do some stuff
   if(someConditionThatRequiresAbort) {
      controller.abort()
   }
}, { signal: controller.signal })
```

lenkan:
@psteniusubi AbortSignal now added to be able to abort the retry prematurely.

psteniusubi:
One more thing. Shouldn't `waitForNotifications` only return unread notifications?

kentbull:
It seems like you would want a separate, explicit function to wait for unread notifications. `waitForUnreadNotifications` would be useful. There is a use case to get all notifications to check whether or not something has been received in the past.

psteniusubi:
I don't think you will ever be waiting for notifications that have already been marked read. You can use notifications().list() to inspect those.

lenkan:
I'll update tomorrow to filter out only unread notifications.

lenkan:
One day late, but updated now. :)

examples/integration-scripts/delegation-multisig.test.ts
```diff
@@ -1,61 +1,47 @@
...
-    let op2 = await client2.oobis().resolve(oobi0.oobis[0], 'delegator');
-    op1 = await waitForOp(client2, op1);
-    op1 = await client2.oobis().resolve(oobi1.oobis[0], 'member1');
-    op1 = await waitForOp(client2, op1);
-    console.log('Member2 resolved 2 OOBIs');
-
-    // First member challenge the other members with a random list of words
-    // List of words should be passed to the other members out of band
-    // The other members should do the same challenge/response flow, not shown here for brevity
-    const words = (await client1.challenges().generate(128)).words;
```
lenkan, line 40: Removed this part because I believe it is already covered by `multisig.tests.ts` and `challenge.test.ts`.

psteniusubi:
Thanks @lenkan. Looks good to me.

m00sey (approved):
lgtm

m00sey:
Merging, multiple reviews, incorporated feedback.

PR #177, opened 2023-12-10
fix challenge.test.ts

nkongsuwan:
When I ran `challenge.test.ts` locally using 
```
TEST_ENVIRONMENT=local npx jest examples/integration-scripts/challenge.test.ts
```

I encountered the following error 
```
 FAIL  examples/integration-scripts/challenge.test.ts (9.743 s)
  ✕ challenge (8165 ms)

  ● challenge

    expect(received).toEqual(expected) // deep equality

    Expected: "bob"
    Received: "Wan"

      93 |     // List Client 1 contacts
      94 |     let contacts1 = await client1.contacts().list();
    > 95 |     expect(contacts1[0].alias).toEqual('bob');
         |                                ^
      96 |     expect(contacts1[0].challenges).toHaveLength(0);
      97 |
      98 |     // Bob responds to Alice challenge

      at Object.<anonymous> (challenge.test.ts:95:32)
```

`contacts1[0]` points to a witness instead of `bob`. I assume that this test passed the automatic test, so I guess the order of elements in `contacts`  is somehow different in the automatic test. Hence, I added a function `findContact()` and a variable `bobContact = findContact(contacts1, 'bob')` to avoid this error.

### DIFF
--- a/examples/integration-scripts/challenge.test.ts
+++ b/examples/integration-scripts/challenge.test.ts
@@ -92,8 +92,9 @@ test('challenge', async () => {
 
     // List Client 1 contacts
     let contacts1 = await client1.contacts().list();
-    expect(contacts1[0].alias).toEqual('bob');
-    expect(contacts1[0].challenges).toHaveLength(0);
+    let bobContact = findContact(contacts1, 'bob');
+    expect(bobContact.alias).toEqual('bob');
+    expect(bobContact.challenges).toHaveLength(0);
 
     // Bob responds to Alice challenge
     await client2.challenges().respond('bob', aid1.i, challenge1_small.words);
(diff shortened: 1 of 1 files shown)


psteniusubi:
Hi @nkongsuwan. 

Looks good to me. You can also lookup contacts by alias if you wish

```
    const list = await client.contacts().list(undefined, 'alias', `^${name}$`);
```

The root cause for the test failing is that there are still some unexpected differences with local and docker environment. This one is caused by differences in https://github.com/WebOfTrust/signify-ts/blob/development/config/keria.json and https://github.com/WebOfTrust/keria/blob/development/scripts/keri/cf/demo-witness-oobis.json

The second Keria config associates witness endpoints with contact alias names "Wan", "Wes" and "Wil".

I'm creating an other PR to fix differences with local and docker environments.

examples/integration-scripts/challenge.test.ts
```diff
@@ -119,5 +120,16 @@ test('challenge', async () => {
 
     // Check Bob's challenge in conctats
     contacts1 = await client1.contacts().list();
-    expect(contacts1[0].challenges[0].authenticated).toEqual(true);
+    bobContact = findContact(contacts1, 'bob');
```
lenkan, line 123:
```suggestion
    bobContact = contacts1.find(contact => contact.name === "bob");
```

Just a suggestion that would work equally well, but can avoid the extra function?

psteniusubi (approved):

lenkan (approved):

PR #178, opened 2023-12-11
Show Python container logs locally

kentbull:
By default in non-interactive mode, when not run as an interactive script, Python doesn't flush the log buffer to stdout or stderr. This means you need to set PYTHONUNBUFFERED=1 to ensure that the output is not buffered, which means it is immediately flushed to stdout or stderr.

In containers this is needed as it makes viewing logs in real-time reliable and consistent.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -1,6 +1,9 @@
 services:
     vlei-server:
         image: gleif/vlei
+        environment:
+            - PYTHONUNBUFFERED=1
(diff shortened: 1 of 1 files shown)


psteniusubi:
Please also add to vlei-server. Looks good otherwise.

I was wondering why logs were not printing on the console. Nice catch thanks!

kentbull:
This has been added to the vlei-server now.

lenkan (approved):

PR #182, opened 2023-12-19
refactor credential tests

lenkan:
I did some work to create some reproduction about credential issuance and revocations in KERIA. When doing that I took the opportunity to do an overhaul of the integration tests of credential issuance from singlesig AIDs.

I refactored the `credentials.test.ts` into smaller "steps" that run in sequence. I also believe I found some overlap in the credential testing, so I think

- single-issuer-holder.test.ts
- single-issuer-chained-credential.test.ts
- request-present.test.ts

are all redundant.

Please give pointers if you think it makes sense to split up the tests like this. It was an attempt to mimic the https://nodejs.org/api/test.html#subtests setup from node.js native test runner. This feature is not available in jest.

This test step separation is also a way to try and setup scenarios more similar to how they would be run in a real world context as you can have local variables and have to think twice before starting to share variables between steps.

Also note the logging output:

Style before this PR:

```
    console.log
      Holder received exchange message with the grant message

      at Object.run (multisig.test.ts:1093:13)

    console.log
      Holder creates and sends admit message

      at Object.run (multisig.test.ts:1104:13)

    console.log
      Member1 received exchange message with the admit response

      at Object.run (multisig.test.ts:1107:13)

    console.log
      Holder holds 0 credential

      at Object.run (multisig.test.ts:1109:13)
```

With this PR (only for this test atm)

```
Finished step - Resolve oobis - 1117ms
Finished step - Create registry - 1153ms
Finished step - issuer can get schemas - 65ms
Finished step - holder can list schemas - 32ms
Finished step - create QVI credential - 1248ms
Finished step - issuer list credentials - 34ms
Finished step - issuer list credentials with filter - 152ms
Finished step - issuer get credential by id - 34ms
Finished step - issuer IPEX grant - 877ms
Finished step - holder IPEX admit - 631ms
Finished step - holder has credential - 12ms
Finished step - holder IPEX present - 452ms
Finished step - verifier receives IPEX grant - 2184ms
Finished step - resolve oobis - 301ms
Finished step - holder create registry for LE credential - 1152ms
Finished step - holder create LE (chained) credential - 1313ms
Finished step - LE credential IPEX grant - 814ms
```


examples/integration-scripts/credentials.test.ts
```diff
@@ -20,318 +23,447 @@ const KLI_WITNESS_DEMO_PREFIXES = [
...
+        assert(holderAgentOOBI.oobis.length >= 1);
+        assert(vfyAgentOOBI.oobis.length >= 1);
+
+        await Promise.all([
+            resolveOobi(issuerClient, QVI_SCHEMA_URL, 'schema'),
+            resolveOobi(holderClient, QVI_SCHEMA_URL, 'schema'),
+            resolveOobi(verifierClient, QVI_SCHEMA_URL, 'schema'),
+            resolveOobi(issuerClient, LE_SCHEMA_URL, 'le-schema'),
+            resolveOobi(holderClient, LE_SCHEMA_URL, 'le-schema'),
+            resolveOobi(verifierClient, LE_SCHEMA_URL, 'le-schema'),
```
kentbull, line 79: I favor having a few explanatory comments in here  to indicate chunks of resolutions, as in Schema OOBIs and Agent OOBIs.

The tests I see as for two audiences:
1. Maintainers
2. KERI learners

Comments primarily help the second group.

lenkan: Good point. Maybe if they are split up into one chunk for each client, it would be more clear?

psteniusubi:
It's good to cleanup this single pretty large integration test. I have a couple of suggestions 

- Use `beforeAll` hooks whenever possible to separate setup code from actual testing code
- Originally I created `getOrCreateIdentifier` and `getOrCreateContact` in addition to `getOrCreateClients` to help with test setup

Have a look at the setup code here
https://github.com/WebOfTrust/signify-ts/blob/2a95620742a5a28cb1d27468e13a1597b3cc08dd/examples/integration-scripts/test-setup-clients.test.ts#L13-L24

This should help remove some clutter.

There's already challenge.test.ts for testing challenge/response, so I think the part could be removed from this.

A couple of smaller details

- WITNESS_AIDS is also available from `TestEnvironment.witnessIds`
- When resolving SCHEMA_OOBI you don't need to specify alias. I've found this useful to keep the contacts list clean
- We should probably add vLEI schema AID constants to resolve-env.ts or other shared component

lenkan:
Thanks for feedback! Will try to get to it during next week.

rodolfomiranda:
I like this refactoring a lot. The test output is cleaner now.

m00sey:
@lenkan any update on the changes?

lenkan:
No sorry. Been AFK for two weeks. Will get back to it tomorrow or Monday. But at first glance I'm not sure I'd want to put stuff in beforeAll or similar since there is only one test.

lenkan:
@m00sey @psteniusubi

Updated the test to use the functions from `test-setup.ts` now. Please take a look when you have the time.

I think the setup code can also take benefit of bypassing console.log to avoid the verbose output from jest. But, if desired, I propose to do that in a separate PR to avoid this PR to explode.

rodolfomiranda (approved):

lenkan:
Did a re-review and found that the 'revoke' step was missing. So added that after your review @rodolfomiranda. Otherwise I think this is good to go.

I will create an issue for adding a step to verify revoked credential from holder side. This is currently commented out in development branch https://github.com/WebOfTrust/signify-ts/blob/2a95620742a5a28cb1d27468e13a1597b3cc08dd/examples/integration-scripts/credentials.test.ts#L311, and just adding the check does not work. So there might be something else to fix there.

PR #188, opened 2024-01-06
add check for revoked credential for holder

### LINKED ISSUE #187: Missing integration test coverage for holding revoked credential
Previously, we had these lines commented out in the credentials test: https://github.com/WebOfTrust/signify-ts/blob/2a95620742a5a28cb1d27468e13a1597b3cc08dd/examples/integration-scripts/credentials.test.ts#L311-L336.

We should introduce coverage for this feature again. However, just adding the same check does not work currently with new IPEX.

lenkan:
First attempt to add coverage to resolve #187 

Created as draft because it does not work at the moment. What are the intended steps here? Another IPEX grant + admit after revocation?

### DIFF
--- a/examples/integration-scripts/credentials.test.ts
+++ b/examples/integration-scripts/credentials.test.ts
@@ -447,4 +447,17 @@ test('single signature credentials', async () => {
 
         assert.equal(issuerCredential.status.s, '1');
     });
+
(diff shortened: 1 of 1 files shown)


pfeairheller:
Close this PR as it is over a month old.  Let's discuss the steps for revocation at the developer's meeting on Thursday

PR #200, opened 2024-01-23
Add an integration test for singlesig vLEI issuance

nkongsuwan:
An integration test for issuance of all types of vLEI credentials.


examples/integration-scripts/singlesig-vlei-issuance.test.ts
```diff
@@ -0,0 +1,580 @@
...
+function createTimestamp() {
+    return new Date().toISOString().replace('Z', '000+00:00');
+}
+
+async function createAid(client: SignifyClient, name: string): Promise<Aid> {
+    const [prefix, oobi] = await getOrCreateIdentifier(client, name);
+    return { name, prefix, oobi };
+}
+
+test('multisig', async function run() {
```
pfeairheller, line 121: A very small nit, but can you change the title for this test to something other than "multisig"?

pfeairheller (changes requested):
Minor update and we can merge this in.

nkongsuwan:
@pfeairheller I have changed the test's name to `singlesig-vlei-issuance` and the test passes local test on my laptop using lastest commits in KERIA. However, multisig.test.ts is failing at the moment which, I think, is unrelated to this test script.

Note: I tried running multisig.test.ts locally as well and it works. No idea why it fails the automatic test on Github.

pfeairheller (approved):
Great, thank you for this test.

PR #203, opened 2024-01-25
chore: add http status information in error message

### LINKED ISSUE #181: HTTP Status code is not included in error information
Currently, the http status code is not included in the error information then signify client receives a non-OK response. See https://github.com/WebOfTrust/signify-ts/blob/d311c1e6c3b4b5674f82cdb00f0d02e5a100bb6f/src/keri/app/clienting.ts#L211-L214.

This means you get a stack trace that does not include any information about the HTTP request and/or response:

```
Error:
   at SignifyClient.fetch...
```

I suggest to include the pathname, the status code, statusText in the error description. The response body can be included as a property on the error. For example

```
new Error(`HTTP ${method} ${pathname} - ${res.status} ${res.statusText}`, { cause: await response.text() });
```

Or create a new error type and add response body as a property.

```
class SignifyClientError extends Error {
    ...Some constructor that create appropriate error message and properties.
}
```

lenkan:
Closes #181

### DIFF
--- a/test/app/clienting.test.ts
+++ b/test/app/clienting.test.ts
@@ -377,4 +377,29 @@ describe('SignifyClient', () => {
             'ELUvZ8aJEHAQE-0nsevyYTP98rBbGJUrTj5an-pCmwrK'
         );
     });
+
+    test('includes HTTP status info in error message', async () => {
+        await libsodium.ready;
+        const bran = '0123456789abcdefghijk';
+        const client = new SignifyClient(url, bran, Tier.low, boot_url);
+
+        await client.connect();
+
+        fetchMock.mockResolvedValue(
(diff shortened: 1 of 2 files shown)


test/app/clienting.test.ts
```diff
@@ -377,4 +377,29 @@ describe('SignifyClient', () => {
...
+        );
+
+        const error = await client
+            .fetch('/somepath', 'GET', undefined)
+            .catch((e) => e);
+
+        assert(error instanceof Error);
+        assert.equal(
+            error.message,
+            'HTTP GET /somepath - 400 Bad Request - Error info'
```
pfeairheller, line 402: This looks good!

pfeairheller (approved):

PR #202, opened 2024-01-25
add test triggering credential revocation crash

lenkan:
Reproduction script for

- https://github.com/WebOfTrust/keria/issues/148
- https://github.com/WebOfTrust/keria/issues/146

PR #207, opened 2024-01-26
Integration test for Multisig delegated vLEI issuance (QVI, LE, ECR)

nkongsuwan:
Add an integration test for multi-signature delegated issuance of QVI, LE, and ECR vLEI credentials:
- The script simulates the process described in the vLEI EGF where QVI has 3 QARs and LE has 3 LARs. 
- Here, GLEIF has 2 GARs that form a GEDA and delegate to the QVI AID. (Technically, the GEDA has 5 External GARs, but the test will get too long with 5 GARs.)
- The ECR credential is issued directly from the LE.

[Edit 18 Feb]: the three problems below seem to have been resolved. 
Currently, there are two problems, one major and one minor:
- [x] Major: The test could not get to the end due to crashes in KERIA after using `client.registries().list()`. This is related to the issues in Signify-TS and KERIA, https://github.com/WebOfTrust/keria/issues/147, https://github.com/WebOfTrust/keria/issues/177, and https://github.com/WebOfTrust/signify-ts/pull/202.
- [x] Minor: KERIA's processing time for credential issuance from a multisig AID to a multisig AID seems to be random, sometimes causing test timeouts.
- [x]  It seems LARs cannot get a credential QARs for some reason.

### DIFF
--- a/examples/integration-scripts/multisig-vlei-issuance.test.ts
+++ b/examples/integration-scripts/multisig-vlei-issuance.test.ts
@@ -1,0 +1,1675 @@
+import { strict as assert } from 'assert';
+import signify, {
+    SignifyClient,
+    Saider,
+    Serder,
+    CredentialSubject,
+    CredentialData,
+    CreateIdentiferArgs,
+    EventResult,
+    randomNonce,
+    Salter,
+} from 'signify-ts';
+import { resolveEnvironment } from './utils/resolve-env';
+import {
+    resolveOobi,
+    waitOperation,
+    waitForNotifications,
+} from './utils/test-util';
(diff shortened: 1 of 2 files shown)


nkongsuwan:
This PR is completed, but it is based on the recent change in PR #222, which hasn't yet been merged.

seriouscoderone:
Why not use `const`?

pfeairheller:
PR #222 has been merged so a rebase here should get this build working.

lenkan (approved):
This LGTM. Only thing is that test time increases significantly, I suppose there is now some overlap in this test and the multisig.test.ts file that can be addressed later.

PR #208, opened 2024-01-30
fix: rotate should use the next threshold as default value for rotate

### LINKED ISSUE #183: Rotate should use next threshold as default value
When building the next state for rotation it uses the current threshold `state.kt` to build the next state.

https://github.com/WebOfTrust/signify-ts/blob/2a95620742a5a28cb1d27468e13a1597b3cc08dd/src/keri/app/aiding.ts#L296

If I'm understanding the rotation logic correctly it should use the `let isith = state.nt` instead?

If this is the only place we build a new state the fix is easy.
But I'm not that up to speed of the entire code base. Are there more places?

When looking at signifypy they seem to have the same bug.

https://github.com/WebOfTrust/signifypy/blob/8f086a60ad15d81a78dfc3653af9e1fa5f775ac1/src/signify/app/aiding.py#L133

However Keri seems to use the next threshold when building the state.

https://github.com/WebOfTrust/keripy/blob/67eece0c00be7318f01bdf507621824e45c03bca/src/keri/app/habbing.py#L1219

lenkan:
Changes the default value for `kt` for the rotation request to use previous state `nt`. 

Closes #183 

I struggled to write a unit test for this scenario, ended up changing the test setup to reduce the dependencies needed to run the test. Introduced the `MockClient` in `aiding.test.ts` so we can simply unit test that we are calling the correct methods on the injected `SignifyClient` rather than having to setup boot and connect replies as well. I much prefer this method for unit testing, if you agree I can try to introduce it to the other modules as well.

Then, only the clienting module tests would verify the signify headers.

### DIFF
--- a/src/keri/app/aiding.ts
+++ b/src/keri/app/aiding.ts
@@ -1,4 +1,3 @@
-import { SignifyClient } from './clienting';
 import { Tier } from '../core/salter';
 import { Algos } from '../core/manager';
 import { incept, interact, reply, rotate } from '../core/eventing';
@@ -7,6 +6,7 @@ import { Tholder } from '../core/tholder';
 import { MtrDex } from '../core/matter';
 import { Serder } from '../core/serder';
 import { parseRangeHeaders } from '../core/httping';
+import { KeyManager } from '../core/keeping';
 
 /** Arguments required to create an identfier */
 export interface CreateIdentiferArgs {
@@ -54,20 +54,26 @@ export interface RotateIdentifierArgs {
 /**
  * Reducing the SignifyClient dependencies used by Identifier class
  */
-export type IdentifierSignifyClient = Pick<
-    SignifyClient,
-    'fetch' | 'pidx' | 'manager'
->;
+export interface IdentifierDeps {
+    fetch(
+        pathname: string,
+        method: string,
+        body: unknown,
+        headers?: Headers
+    ): Promise<Response>;
+    pidx: number;
+    manager: KeyManager | null;
+}
 
 /** Identifier */
 export class Identifier {
-    public client: IdentifierSignifyClient;
+    public client: IdentifierDeps;
 
     /**
      * Identifier
-     * @param {IdentifierSignifyClient} client
+     * @param {IdentifierDeps} client
      */
-    constructor(client: IdentifierSignifyClient) {
+    constructor(client: IdentifierDeps) {
         this.client = client;
     }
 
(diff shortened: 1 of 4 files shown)


src/keri/app/aiding.ts
```diff
@@ -51,14 +51,23 @@ export interface RotateIdentifierArgs {
...
 
+/**
+ * Reducing the SignifyClient dependencies used by Identifier class
+ */
+export type IdentifierSignifyClient = Pick<
+    SignifyClient,
+    'fetch' | 'pidx' | 'manager'
+>;
+
 /** Identifier */
```
pfeairheller, line 62: What does this do?

lenkan: TLDR: Declares the dependencies of the Identifier class as a subset of the type that is derived from SignifyClient.

This is a way to declare the dependencies of the `Identifier` class. It only depends on the `fetch`, `pidx` and `manager` properties of SignifyClient, all of which are easy to create test versions of and inject into the class for testing. There are multiple patterns to achieve this, this was the first that came to mind so I just went with it.

Other alternatives are:

- Type casting in aiding.test.ts, leaving aiding.ts as is:

```ts
// aiding.ts
class Identifier {
  constructor(client: SignifyClient) { }
}

// aiding.test.ts

class MockClient {
   fetch = jest.fn();
   pidx = 0;
   manager: KeyManager;
   
   identifiers() {
      return new Identifier(this as unknown as SignifyClient)
   }
}
```

Declaring the dependency explicitly in aiding.ts and have the MockClient implement the same dependency.

```ts
// aiding.ts
interface IdentifierDeps {
   fetch(url: string, method: string, body: unknown, header?: Headers): Promise<Response>;
   pidx: number;
   manager: KeyManager;
}

class Identifier {
  constructor(client: IdentifierDeps) { }
}

// aiding.test.ts
class MockClient implements IdentifierDeps {
   fetch = jest.fn();
   pidx = 0;
   manager: KeyManager;
   
   identifiers() {
      return new Identifier(this)
   }
}
```

If would also be optional to add the `implements IdentifierDeps` to `SignifyClient` class. But it already implicitly implements the interface so it is redundant. I can also see a case in the near future where the `fetch` method is broken out from SignifyClient to make things a bit more easily testable.

Since you raised the question and I am looking at this again, I prefer the last pattern with `IdentifierDeps`. I don't mind adding this to all other client classes if people think this is a good pattern going forward.

m00sey (approved):
lgtm

PR #212, opened 2024-02-09
Multisig inception utils for testing

lenkan:
When I was creating the reproduction script for https://github.com/WebOfTrust/keria/issues/189 I extract a couple of utility functions for multisig inceptions. I thought it might be useful to include them in the multisig tests to make it a bit easier to comprehend the tests.

In the future, I believe the "accept" type of function for multisig operations would be useful to include in the exposed library.

### DIFF
--- a/examples/integration-scripts/delegation-multisig.test.ts
+++ b/examples/integration-scripts/delegation-multisig.test.ts
@@ -1,6 +1,7 @@
 import { strict as assert } from 'assert';
 import signify from 'signify-ts';
 import {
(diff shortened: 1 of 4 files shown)


lenkan:
Would love some feedback if people think this abstraction is good. In that case I could add some more "startMultisigXXX"/"acceptMultisigXXX" functions and refactor the tests to make use of them. Then we can consider moving them into the library.

pfeairheller (approved):

Issue #138, opened 2023-11-16
Review seqner usage (decimal vs hex)


rodolfomiranda:
Review if we are using/calling the seqner correctly in a bunch of places in the code regarding sequence number being decimal or hexadecimal.  It'd be helpful to add tests that run over 10 key events.

I assigned the issue to myself as a reminder.

lenkan:
@rodolfomiranda I encountered this I think. Currently we do `const sn = Number(state.s)` which only works for `state.s < 10`. If `state.s` is expected to be in hexadecimal without the `0x` prefix, we should do this instead: `const sn = parseInt(state.s, 16)`. I have a fix on my fork, but it contains some other fixes that I need to branch out as well. I probably won't get to it until monday.

psteniusubi:
Keripy definition for numbers from [Number](https://github.com/WebOfTrust/keripy/blob/86c50b466f3fb6b8a55615ce40ba1cc31d10b749/src/keri/core/coring.py#L1572C15-L1572C15):

>         Parameters:
>             num (int | str | None): non-negative int number or hex str of int
>                 number or 0 if None
>             numh (str):  string equivalent of non-negative int number
> 
>         Note: int("0xab", 16) is also valid since int recognizes 0x hex prefix

I think there's a logic error in CesrNumber of signify-ts. It does not match the logic of Number in keripy

https://github.com/WebOfTrust/signify-ts/blob/421232df0dd52418a917ceff7f2b7d3e398d1d29/src/keri/core/number.ts#L17-L23

vs

https://github.com/WebOfTrust/keripy/blob/86c50b466f3fb6b8a55615ce40ba1cc31d10b749/src/keri/core/coring.py#L1582-L1597

Also there are probably issues using `parseInt` with large integers (bigger than _safe integer_ or 2^53)

CesrNumber should be fixed. Review code for uses of `parseInt(input)` or `Number(input)`. Where needed replace with

`const result = new CesrNumber({}, input, undefined).num`

daidoji:
Has this been fixed?  It looks like it on main (or at least lots of changes have happened since this issue was filed).  https://github.com/WebOfTrust/signify-ts/blob/main/src/keri/core/number.ts

iFergal:
[...]

PR #216, opened 2024-02-14
fix sn to int parsing to use base 16

lenkan:
The sequence number is expected to be a hex string without the preceeding `0x`. So for example sequence number 10 is simply `a`. This causes issues with previous implementation because `Number("a") => NaN` this fixes the issue by specifying the base when parsing.

Another potential fix could have been to add a `0x` prefix server side... But that probably has more implications that I am not aware of.

It would be good to add some unit tests with that verifies the behavour with a sequence number > 10.

There is a related issue where I wrote about this a few months ago: https://github.com/WebOfTrust/signify-ts/issues/138. Though I think the issue is broader than this particular fix. I simply searched the codebase for usage of the `Number` constructor and replaced them where it looked like it needed replacing.

### DIFF
--- a/src/keri/core/number.ts
+++ b/src/keri/core/number.ts
@@ -40,7 +40,7 @@ export class CesrNumber extends Matter {
             // make huge version of code
             code = code = NumDex.Huge;
         } else {
-            throw new Error('Invalid num = {num}, too large to encode.');
+            throw new Error(`Invalid num = ${num}, too large to encode.`);
         }
 
         raw = intToBytes(_num, Matter._rawSize(code));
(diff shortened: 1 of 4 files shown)


kentbull:
I'm thinking [this](https://github.com/WebOfTrust/signify-ts/blob/cedee3ac163eb7111896a7479ba381a7f6d0649d/src/keri/core/eventing.ts#L70) `Number(sn)` call likely also needs to be using `parseInt(sn, 16)`.

src/keri/core/number.ts
```diff
@@ -40,7 +40,7 @@ export class CesrNumber extends Matter {
             // make huge version of code
             code = code = NumDex.Huge;
         } else {
-            throw new Error('Invalid num = {num}, too large to encode.');
+            throw new Error(`Invalid num = ${num}, too large to encode.`);
```
pfeairheller, line 43: Good catch!

pfeairheller (approved):
Thanks for the new tests

PR #217, opened 2024-02-14
Align passcode generation with `kli salt`

kentbull:
Using 32 bytes rather than 16 bytes aligns SignifyTS passcode generation with KERIpy's `kli salt` command.

The generatePasscode function from SignifyTS has generated some passcodes that don't end up working for me.

```typescript
export function randomPasscode(): string {
    const raw = libsodium.randombytes_buf(16);
    const salter = new Salter({ raw: raw });

    return salter.qb64.substring(2);
}
```

From the kli salt command it is similar, though not quite the same:
```python
print(coring.Salter(raw=pysodium.randombytes(pysodium.crypto_sign_SEEDBYTES)).qb64)
```

For whatever reason the libsodium.randombytes_buf(16)  in Typescript doesn't produce the same results as pysodium.randombytes(pysodium.crypto_sign_SEEDBYTES)

I have seen problems with any random 21 character string which is why I started trying to constrain it with functions from Signify.

I found the following it in the Libsodium Github Repo [here](https://github.com/jedisct1/libsodium/blob/b564794eddda79d54374144653c1f4e05ae08a64/src/libsodium/include/sodium/crypto_sign_ed25519.h#L26). In the C code 
```c
// in libsodium/src/libsodium/include/sodium
/crypto_sign.h
#define crypto_sign_SEEDBYTES crypto_sign_ed25519_SEEDBYTES

// which references the following in
// src/libsodium/include/sodium/crypto_sign_ed25519.h
#define crypto_sign_ed25519_SEEDBYTES 32U
```

Looks like the Python impl uses 32 bytes and SignifyTS is currently using 16 bytes.

### DIFF
--- a/src/keri/app/coring.ts
+++ b/src/keri/app/coring.ts
@@ -4,7 +4,7 @@ import { Salter } from '../core/salter';
 import { Matter, MtrDex } from '../core/matter';
 
 export function randomPasscode(): string {
-    const raw = libsodium.randombytes_buf(16);
+    const raw = libsodium.randombytes_buf(32);
     const salter = new Salter({ raw: raw });
 
     return salter.qb64.substring(2);


SmithSamuelM:
There is some confusion here. A true random salt only needs 16 bytes of entropy to be cryptographically strong. The private keys for libsodium are called seeds in libsodium. They are used to generate public keys via a scalar (seed) multiplication of an ECC point. The strength of ECC scalar multiplication is only the square root of the size of the scalar (seed). So in order to have cryptographic strength of 128 bits of entropy the scalar multiplication must be 256 bits. Hence the reason for libsodium seeds being 256 bits (32 bytes). When a salt is not being used as a seed for scalar multiplication but is used to inject entropy into a hash then it only needs to be 16 bytes. The hash itself needs to be 32 bytes for the same reason. The cryptographic strength of the hash is only the square root of the size of the hash. So a 32 byte hash has only 16 bytes of cryptographic strength. So in the interest of economy, there is little to no benefit to using random Salts bigger than 16 bytes. When those Salts are used to generate hashes that that are 32 bytes or seeds that are 32 bytes.

pfeairheller (approved):

kentbull:
Thank you for the detailed explanation with distinctions between the salt and seed. This makes sense to me now. I'll close this issue.
If I understand correctly what you've said then for consistency we may want to change the `kli salt` command in KERIpy to use only 16 bits of entropy.

PR #218, opened 2024-02-15
fix: remove urlsafe-base64 dependency

lenkan:
In PR #214, I missed that the urlsafe-base64 dependency used the global `Buffer` object internally. I already had some polyfills added when I tested the change so I missed that we still required the polyfill.

There are a few solutions to this.

- Re-introduce the polyfill but also set `globalThis.Buffer = import("buffer").Buffer`.
- Stop using "Buffer" completely, just use Uint8Array which is available in both browser and node.js runtime.
- Implement base64url encode/decode using the imported `buffer` buffer module.

I think the first one is a hack and an anti-pattern to taint global namespace in a library. The second one is a bit too much work for a quick fix. The last one is a good compromise in my opinion, so I did that. Ideally, the buffer module should support base64 url encode/decode but the PR for that has not been merged for some time: https://github.com/feross/buffer/pull/314.

I took some code and test cases from the node.js ecosystem and replaced the urlsafe-base64 dependency with an internal module that uses the buffer package.

Let me know what you think. Can also go with option 1 if you prefer. But a fix is needed to avoid forcing users to polyfill for themselves.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -17,8 +17,7 @@
         "ecdsa-secp256r1": "^1.3.3",
         "libsodium-wrappers-sumo": "^0.7.9",
         "mathjs": "^12.4.0",
-        "structured-headers": "^0.5.0",
-        "urlsafe-base64": "^1.0.0"
+        "structured-headers": "^0.5.0"
       },
       "devDependencies": {
         "@mermaid-js/mermaid-cli": "^10.3.0",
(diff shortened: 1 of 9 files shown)


psteniusubi:
Less dependencies is better in my opinion. Ultimately I think we should get rid of Buffer completely.

pfeairheller (approved):
I think this is a good compromise.  I doubt there will be any innovations in Base64 encoding/decoding we'll have to keep up with!

Issue #221, opened 2024-02-18
No support for multisig issuance of private ACDCs


nkongsuwan:
Currently, [`client.credentials().issue(kargsIss)`](https://github.com/WebOfTrust/signify-ts/blob/development/src/keri/app/credentialing.ts#L181) does not support multisig issuance of private ACDCs. 

The reason is that [`IssueCredentialArgs`](https://github.com/WebOfTrust/signify-ts/blob/development/src/keri/app/credentialing.ts#L39) doesn't allow arbitrary salts `u` as input to the function `issue()`. As a result, when members of a multisig group try to issue a private ACDC, they will get inconsistent salts, `acdc.u` and `acdc.a.u`.

I suggest changing the interface `IssueCredentialArgs` to include inputs for the top-level salt and the one inside `a`, e.g., `topLevelSalt?: boolean | string` and `attributeSalt?: boolean | string`.

This issue is blocking PR #207  for the issuance of an ECR vLEI credential.

psteniusubi:
Hi @nkongsuwan. I have run into this same issue.

I have not created a PR yet, but I have a quick and dirty fix that adds two parameters to `IssueCredentialArgs`

```
$ git stash show -p stash@{1}
diff --git a/src/keri/app/credentialing.ts b/src/keri/app/credentialing.ts
index 058ccdd..bd8feb2 100644
--- a/src/keri/app/credentialing.ts
+++ b/src/keri/app/credentialing.ts
@@ -81,6 +81,8 @@ export interface IssueCredentialArgs {
      * Flag to issue a credential with privacy preserving features
      */
     privacy?: boolean;
+    subjectSalt?: string;
+    acdcSalt?: string;
 }

 export interface IssueCredentialResult {
@@ -196,7 +198,9 @@ export class Credentials {

         const [, subject] = Saider.saidify({
             d: '',
-            u: args.privacy ? new Salter({}).qb64 : undefined,
+            u: args.privacy
+                ? args.subjectSalt ?? new Salter({}).qb64
+                : undefined,
             i: args.recipient,
             dt: dt,
             ...args.data,
@@ -205,7 +209,7 @@ export class Credentials {
         const [, acdc] = Saider.saidify({
             v: versify(Ident.ACDC, undefined, Serials.JSON, 0),
             d: '',
-            u: args.privacy ? new Salter({}).qb64 : undefined,
+            u: args.privacy ? args.acdcSalt ?? new Salter({}).qb64 : undefined,
[...]

PR #222, opened 2024-02-19
fix: cannot set salts for credential creation (suggestion)

lenkan:
This is my suggestion on how to solve #221 .

I know this is a breaking change, but I don't think we need the additional interface for IssueCredentialArgs. I think it just adds confusion with the additional mapping of properties. Same for the other operations such as create identifier, rotate etc.

### DIFF
--- a/examples/integration-scripts/multisig-holder.test.ts
+++ b/examples/integration-scripts/multisig-holder.test.ts
@@ -1,10 +1,5 @@
 import { strict as assert } from 'assert';
-import signify, {
-    SignifyClient,
-    IssueCredentialArgs,
-    Operation,
-    CredentialData,
-} from 'signify-ts';
+import signify, { SignifyClient, Operation, CredentialData } from 'signify-ts';
 import { resolveEnvironment } from './utils/resolve-env';
 import {
     assertOperations,
--- a/examples/integration-scripts/multisig-holder.test.ts
+++ b/examples/integration-scripts/multisig-holder.test.ts
@@ -3,6 +3,7 @@ import signify, {
     SignifyClient,
     IssueCredentialArgs,
     Operation,
+    CredentialData,
 } from 'signify-ts';
 import { resolveEnvironment } from './utils/resolve-env';
 import {
@@ -354,12 +355,11 @@ test('multisig', async function run() {
 
     console.log(`Issuer starting credential issuance to holder...`);
     const registires = await client3.registries().list('issuer');
-    await issueCredential(client3, {
-        issuerName: 'issuer',
-        registryId: registires[0].regk,
-        schemaId: SCHEMA_SAID,
-        recipient: holderAid['prefix'],
-        data: {
+    await issueCredential(client3, 'issuer', {
+        ri: registires[0].regk,
+        s: SCHEMA_SAID,
+        a: {
+            i: holderAid['prefix'],
             LEI: '5493001KJTIIGC8Y1R17',
         },
     });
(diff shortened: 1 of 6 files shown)


psteniusubi:
It's good if the API can accept exn messages directly, without property mapping.

The API probably becomes harder to use with the compact one letter property names. Need to make sure all required properties are declared and have some reasonable documentation.

examples/integration-scripts/multisig-holder.test.ts
```diff
@@ -354,12 +350,11 @@ test('multisig', async function run() {
 
     console.log(`Issuer starting credential issuance to holder...`);
     const registires = await client3.registries().list('issuer');
-    await issueCredential(client3, {
-        issuerName: 'issuer',
-        registryId: registires[0].regk,
-        schemaId: SCHEMA_SAID,
-        recipient: holderAid['prefix'],
-        data: {
+    await issueCredential(client3, 'issuer', {
```
seriouscoderone, line 353: The readability of code is paramount. 

We should fix this at the serialization level.

One option: 
https://www.npmjs.com/package/typescript-json-serializer

```
@JsonObject()
export class Animal extends LivingBeing {
    @JsonProperty() name: string | undefined;
    @JsonProperty() birthDate: Date;
    @JsonProperty() numberOfPaws: number;
    @JsonProperty() gender: Gender;

    // Enum value (string)
    @JsonProperty() status: Status;

    // Specify the property name of json property if needed
    @JsonProperty('childrenIdentifiers')
    childrenIds: Array<number>;

    constructor(name: string | undefined) {
        super();
        this.name = name;
    }

}
```

seriouscoderone: This uses decorators on the objects. And leaves the code readable.

lenkan:
I just listened in on the dev call and wanted to provide some comments.

It seems like this change sparked some discussion on which direction to take regarding the API. I am voting for going towards a signify API that as closely as possible resembles the API of KERIA. The additional property mapping just is just another layer of indirection that in itself requires documentation. Some benefits of the approach:

- Documentation of the signify-ts API can be derived or implied from KERIA API documentation.
- Typescript types for request and response object shapes can probably be completely generated by openapi tooling (once Openapi specs for keria is complete).
- Less issues of the type #221 (there has previously been more issues like this)

Also, for this particular API of issuing credential, a user of this library could add their own functions for ACDC creation according to which ever schema they want to use, like this.

```typescript
function createQviCredentialData(args: QviCredentialData): CredentialData {
    return {
        ri: args.registryId,
        i: args.issuerPrefix,
        s: "EBf...",
        a: { LEI: args.LEI, i: args.recipientPrefix }
    }
}

const data = createQviCredentialData({ registryId, issuerPrefix: "abc", recipientPrefix: "def", LEI: "123" })
await credentials().issue("issuer", data)
```

As @psteniusubi mentioned as well, the fact that the API can accept the same shape of objects that is passed in exchange messages is a huge plus in my opinion, as can be seen in this change: https://github.com/WebOfTrust/signify-ts/pull/222/commits/7022f95e7056862e1ec5a66d85938732656f70a1#diff-6d0716ef2bd3e3e9494efe4b175ef940675ac2c0615c4d98172c7c5f69671c9aR906.

If we decide to go for this approach, we should _probably_ plan to change the rest of the API to do follow the same strategy and then strictly adhere to it going forward.

I do see the argument against it as well though, using more descriptive property names is helpful, but the benefits mentioned are just much greater in my opinion.

psteniusubi:
@lenkan - I agree with the suggestion to move forward with compact labels.

The [Keri spec](https://trustoverip.github.io/tswg-keri-specification/#keri-data-structure-format) already has definitions for message types and labels. (I could not immediately find all message types, for example "iss" and "vcp" are missing.)

As you suggest it is relatively easy to build a higher lever abstraction on top of the lower level Signify API. One could even implement a backwards compatibility layer.

There is at least one catch. Because Keri sets strict rules on [ordering](https://trustoverip.github.io/tswg-keri-specification/#special-label-ordering-requirements) of the fields, the Signify API should still internally re-build all messages. I don't think the API can internally use the spread `...` operator. For messages that originate from the API client ordering is not known, or there could exist unknown fields.

Should this design change apply to signifypy as well?

pfeairheller (approved):

PR #223, opened 2024-02-24
simplify prettier setup

lenkan:
This setup uses an "opt-out" approach through a .prettierignore file rather than the "opt-in" approach.

Some benefits:

- All files are autoformatted unless explicitly ignored.
- Works cross platform (not sure if the shell expansion approach works on windows)
- Same ignore pattern will also be picked up by editors. Previously, when using a prettier plugin, it would auto format package.json, even though we have previously excluded it from the formatting.
  - https://marketplace.visualstudio.com/items?itemName=esbenp.prettier-vscode
  - https://www.jetbrains.com/help/idea/prettier.html#ws_prettier_install
  
I added package.json and the workflow files to `.prettierignore` because they currently use `tab-width: 2` instead of 4 like the prettier config says.


m00sey (approved):
this lgtm, but I am not digesting it daily.

PR #228, opened 2024-02-28
Remove workspace from package.json

s-a-tanjim:
The problem is, when we try to build using Yarn, it gives an error `Workspaces can only be enabled in private projects`. Yarn intentionally does this. You can find the issue here: https://github.com/yarnpkg/yarn/issues/8580.
And in [this](https://classic.yarnpkg.com/lang/en/docs/workspaces/) doc, they said,
`Note that the private: true is required! Workspaces are not meant to be published, so we’ve added this safety measure to make sure that nothing can accidentally expose them.`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 .DS_Store
 node_modules
 dist
-
+yarn.lock*
 
 # IntelliJ Project Files
 .idea
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,13 @@
 {
   "name": "signify-ts",
-  "version": "0.1.1",
+  "version": "0.2.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "signify-ts",
-      "version": "0.1.1",
+      "version": "0.2.1",
       "license": "Apache-2.0",
-      "workspaces": [
-        "examples/*"
-      ],
       "dependencies": {
         "@noble/hashes": "^1.3.2",
         "buffer": "^6.0.3",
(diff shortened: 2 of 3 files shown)


s-a-tanjim:
The problem is not with this project. We have another project that has the signify-ts dependency. We use Yarn there. If we use npmjs (ie: `"signify-ts": "0.2.1"`) it doesn't create the problem. But when we use github url (ie: `"signify-ts": "https://github.com/WebOfTrust/signify-ts.git"`), then Yarn doesn't let us build the project. We have to use github url because npmjs doesn't get updates frequently. So if we can't solve this issue here, maybe we need to migrate our project into npm.

lenkan:
I haven't used yarn in many years so I don't have any other suggestions. Imo, this is a yarn problem. But perhaps another solution is to start publishing this package on every commit. There is an issue for it https://github.com/WebOfTrust/signify-ts/issues/127.

Meanwhile you could publish a scoped version of the package from your own fork.

lenkan:
@s-a-tanjim  I also noted now that the issue you linked is for yarn v1. I thought that project was dead many years ago? Have you tried https://github.com/yarnpkg/berry ?

s-a-tanjim:
@lenkan Yeah I tried with Yarn V4 [berry](https://github.com/yarnpkg/berry). It works! But I don't know why they still make the Yarn V1 the default version. And this https://github.com/WebOfTrust/signify-ts/issues/127 issue will make our life a lot easier! The main idea behind this PR was `Workspaces are not meant to be published`. But I understand the need for this in `signify-ts` project. So I'm closing this PR.

Issue #230, opened 2024-03-07
signature-params component of signature-base is not serialized


da-tychinin:
As far as I understand the http messages signature process follows the [HTTP Messages Signatures RFC](https://datatracker.ietf.org/doc/html/rfc9421)

The algorithm to build a signature-base for signing/verification contains a discrepancy with the HTTP messages signatures RFC. The RFC says that signature-params component  must be canonicalized as _Inner List Structured Field values with parameters_. The current implementation is missing [parameters serialization](https://www.rfc-editor.org/rfc/rfc8941#section-4.1.1.2). Here is a link to the exact line of code: https://github.com/WebOfTrust/signify-ts/blob/e2b2f6e3d9ecd4d1ae22e93ab82ded4ef366312b/src/keri/core/httping.ts#L103

The fix is to update the code with signature-params serialization. That will include changes for signify-ts, signifypy and any other implementations that are involved into signature generation/validation.

rodolfomiranda:
Great catch.

da-tychinin:
Here is the signature-base string example before and after the fix. Please take a look at the @signature-params component and DQUOTES in there:
Before:
```
"signify-resource": EWJkQCFvKuyxZi582yJPb0wcwuW3VXmFNuvbQuBpgmIs
"@method": POST
"@path": /signify
"signify-timestamp": 2022-09-24T00:05:48.196795+00:00
"@signature-params": ("signify-resource" "@method" "@path" "signify-timestamp");created=1609459200;keyid="DN54yRad_BTqgZYUSi_NthRBQrxSnqQdJXWI5UHcGOQt";alg="ed25519"
```
After:
```
"signify-resource": EWJkQCFvKuyxZi582yJPb0wcwuW3VXmFNuvbQuBpgmIs
"@method": POST
"@path": /signify
"signify-timestamp": 2022-09-24T00:05:48.196795+00:00
"@signature-params: (signify-resource @method @path signify-timestamp);created=1609459200;keyid=DN54yRad_BTqgZYUSi_NthRBQrxSnqQdJXWI5UHcGOQt;alg=ed25519"
```

rodolfomiranda:
the last quote is misplaced, right?

da-tychinin:
@rodolfomiranda sorry, I've misplaced the examples in the previous comments. I've edited the comment above and now they are correct
about double quotes - there are a few missing:
* `@signature-params` component name should be wrapped into double quote - in the example the last one is missing
[...]

PR #231, opened 2024-03-08
230 signature base serialization fix

da-tychinin:
Issue: https://github.com/WebOfTrust/signify-ts/issues/230
* moved all signature input generation from authing to httping
* added optional authority parameter for authing methods.
* removed Signer calls from httping. Moved them to the upper level - authing
* refactor siginput and desiginput method to make them reflect one another
* desiginput return value changed to map by signature label. According to the RFC the labels must be unique
* added sigbase method to httping. It is now used in both sign and verify authing functions
* authing.test.ts: mock httping methods.
* added RFC test cases

### DIFF
--- a/src/keri/core/authing.ts
+++ b/src/keri/core/authing.ts
@@ -1,11 +1,14 @@
 import { Signer } from './signer';
 import { Verfer } from './verfer';
-import { desiginput, normalize, siginput } from './httping';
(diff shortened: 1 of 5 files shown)


da-tychinin:
I've fixed the error TS5097 for "build, lint and test" stages. Should work fine now. Integration test will fail because of the different signature-base serialization

Issue #204, opened 2024-01-26
Handling of special characters and white space in alias names


lenkan:
When creating an identifier with "special" characters, for example spaces or the "ö" letter used in Swedish and Finnish:

```js
    const alias = 'ö';
    const result = await client.identifiers().create(alias, {});
    await waitOperation(client, await result.op());
```

You will get an error when signify-ts is doing the fetch call:

```
    TypeError: fetch failed

      204 |             });
      205 |         }
    > 206 |         const res = await fetch(this.url + path, {
          |                     ^
      207 |             method: method,
      208 |             body: _body,
      209 |             headers: final_headers,

      at async SignifyClient.fetch (../../src/keri/app/clienting.ts:206:21)
      at async EventResult.op (../../src/keri/app/aiding.ts:467:21)
      at async Object.<anonymous> (identifier-name-with-special-chars.test.ts:8:33)

    Cause:
    RequestContentLengthMismatchError: Request body length does not match content-length header
```

I also experimented with URL encoding the alias:

```js
    const alias = encodeURIComponent('ö');
    const result = await client.identifiers().create(alias, {});
    await waitOperation(client, await result.op());

    // Call below fails with a 401 Unauthorized
    const id = await client.identifiers().get(alias);
```

Then the creation succeeds, but a subsequent `identifiers().get(alias)` results in a 401 Unauthorized from keria. The same thing happens when creating an alias with spaces:

```js
    const alias = "a b";
    const result = await client.identifiers().create(alias, {});
    await waitOperation(client, await result.op());

    
    // Call below fails with a 401 Unauthorized
    const id = await client.identifiers().get(alias);
```

I have put the reproduction scripts here: https://github.com/lenkan/signify-ts/blob/2d21adae1f3706f7829566d9af10f0f57d8741dc/examples/integration-scripts/identifier-name-with-special-chars.test.ts

Not sure if this issue should be in signify-ts or in keria. Depending on the route taken, it could affect both repos. I am totally fine with rejecting special characters in identifier names altogether, but then we should probably do some request input validation in keria to reject any invalid requests with a 4xx error.

2byrds:
From our keri dev meeting:
What is the expected behavior for alias names?
Should we be able to support all utf-8?
Localisation is a broader discussion.
The alias is in the path and is encoded.
A fix was added where the content length header was being manually calculated instead of by the fetch.
Probably it was unauthorized because the path changed and the verification of the signature failed.
[...]

PR #233, opened 2024-03-14
URL enconding of AID alias name in rest calls

rodolfomiranda:
This PR fix issue #204 **Handling of special characters and white spaces in alias names** by adding a missing url encoding and the correct `content-length` calculation. 

Without changes in KERIA it will fix the spaces and some special characters such as @ and $, but not other special character such as the `ö` letter used in Swedish.
In KERIA, it seems that Falcon modifies the `req.path` when passing it to the middleware that check signatures. Need further investigation.

### DIFF
--- a/src/keri/app/clienting.ts
+++ b/src/keri/app/clienting.ts
@@ -183,7 +183,7 @@ export class SignifyClient {
 
         const _body = method == 'GET' ? null : JSON.stringify(data);
         if (_body !== null) {
-            headers.set('Content-Length', String(_body.length));
+            headers.set('Content-Length', String((new TextEncoder().encode(_body)).length));
         }
         if (this.authn) {
             signed_headers = this.authn.sign(
--- a/src/keri/app/aiding.ts
+++ b/src/keri/app/aiding.ts
@@ -112,7 +112,7 @@ export class Identifier {
      * @returns {Promise<any>} A promise to the identifier information
      */
     async get(name: string): Promise<any> {
-        const path = `/identifiers/${name}`;
+        const path = `/identifiers/${encodeURIComponent(name)}`;
         const data = null;
         const method = 'GET';
         const res = await this.client.fetch(path, method, data);


src/keri/app/clienting.ts
```diff
@@ -183,7 +183,10 @@ export class SignifyClient {
 
         const _body = method == 'GET' ? null : JSON.stringify(data);
         if (_body !== null) {
-            headers.set('Content-Length', String(_body.length));
+            headers.set(
```
lenkan, line 186: I think setting this header is redundant unless signify has any special requirements for it. I believe `fetch` will automatically set the content-length header for you.

lenkan (commented):
Nice! It would be nice with a unit test for the path encoding. Should be fairly straightforward in aiding.test.ts

src/keri/app/clienting.ts
rodolfomiranda: I think you are right. I'll do a test and remove that line if the header is automatically added.

rodolfomiranda: I validated that `fetch` adds it. I'm removing the lines

rodolfomiranda:
> Nice! It would be nice with a unit test for the path encoding. Should be fairly straightforward in aiding.test.ts

added

rodolfomiranda:
You are right. I updated the code.

test/app/aiding.test.ts
```diff
@@ -99,6 +99,15 @@ describe('Aiding', () => {
         assert.deepEqual(lastCall.body.salty.transferable, true);
     });
 
+    it('Can create salty identifiers with special characters in the name', async () => {
```
lenkan, line 102:
```suggestion
    it('Can get identifiers with special characters in the name', async () => {
```

lenkan (approved):
LGTM

lenkan:
@rodolfomiranda Should we merge this?

rodolfomiranda:
> @rodolfomiranda Should we merge this?

Yes!

PR #234, opened 2024-03-21
feat: single-sig IPEX apply, offer, agree

iFergal:
To accompany https://github.com/WebOfTrust/keria/pull/198.


src/keri/app/credentialing.ts
```diff
@@ -96,6 +96,85 @@ export interface RevokeCredentialResult {
...
+
+    /**
+     * Message to send
+     */
+    message?: string;
+
+    /**
+     * SAID of schema to apply for
+     */
+    schema: string;
```
iFergal, line 118: Unsure looking at the spec if we should also be able to pass a full schema, and not just the SAID.

src/keri/app/credentialing.ts
```diff
@@ -713,6 +792,139 @@ export class Ipex {
...
+    /**
+     * Create an IPEX apply EXN message
+     */
+    async apply(args: IpexApplyArgs): Promise<[Serder, string[], string]> {
+        const hab = await this.client.identifiers().get(args.senderName);
+        const data = {
+            m: args.message ?? '',
+            s: args.schema,
+            a: args.attributes ?? {},
+            i: args.recipient,
```
iFergal, line 804: Not sure on when `i` should be part of the attributes here. It is for grant and not for apply. From what I remember when checking out keripy, it seems to auto add `i` if an exn message is created there.

lenkan: @iFergal Do you have a link to the lines of code in keripy where this is?

src/keri/app/credentialing.ts
lenkan: I suppose that can be added later if needed.

src/keri/app/credentialing.ts
iFergal: @lenkan Sure - https://github.com/WebOfTrust/keripy/blob/development/src/keri/peer/exchanging.py#L333

lenkan: I had a look here and it is being done in here as well: https://github.com/WebOfTrust/signify-ts/blob/ac6dac44cec401f6d95df19f3e2b475d9367995a/src/keri/app/exchanging.ts#L194-L214

I think this is more explicit, we can amend the interface for the payload parameter in `createExchangeMessage` to be something like

```
interface ExchangeMessageData {
   [key: string]: unknown,
   /**
     * The recipient of the message
     */
   i: string
}
```

that way there is only one way to specify the recipient and we can remove the recipient parameter. This would be more in line with the changes I proposed in https://github.com/WebOfTrust/signify-ts/pull/222

iFergal: Ah I see, though that line of code is never being hit for IPEX since `recipient` is always passed as `undefined` to `createExchangeMessage`. The only other place that `createExchangeMessage` is called is from the `send` method of exchanges, which also doesn't pass the recipient.

Maybe @pfeairheller can shed more light on when it should be passed as an attribute too (at least for IPEX).

But agree on the interface tidy up, makes sense to me.

iFergal: @lenkan I think the pipeline should pass now with a re-run (I can't trigger). Not sure if I need to make any changes here regarding the recipient first but it would be good to get this merged at some point.

lenkan: I could not trigger a run for this PR either. Perhaps a new force push is enough. I see https://hub.docker.com/r/weboftrust/keria/tags has not been updated for 21 days.

iFergal: I think `matrix.keria-version` needs to be specifically set to the dev tag, and not `latest`. For me this pulls `0.1.3` which seems outdated.

If we're OK with pinning for the tests I can push that change too, but we'll have to update it every time.

lenkan: I am OK with it. Although, it essentially means that we are dropping support for the released version of KERIA, right? I would also change the fallback value in docker-compose.yaml. You can see the changes in the build yaml and docker-compose.yaml here: https://github.com/WebOfTrust/signify-ts/pull/252/files

When 0.2.0-dev* gets promoted to latest we can change back.

lenkan: I am also totally open to get rid the test matrix if it is too awkward to change the targeted version in two places. I thought we would run it against multiple versions of keria at first, but I don't think we are going to do that for some time.

iFergal: Hmm, I think it makes sense for the pipeline to _always_ point at the dev version, even if this one gets promoted. Because it will just happen again.

Maybe the docker compose can still default to `latest` but the CICD overrides using the matrix to `0.2.0-dev*` and is updated as people require it.

iFergal: @lenkan We discussed this in the dev call today, and I also can see you have #252 open that also does this. I've pushed the change for the CICD and compose file only to see if it will pass now. Not sure if additionally all of the other changes you have in draft are required too or not with this new version.

cc @rodolfomiranda

iFergal: @lenkan Do you think that PR will merge this week or shall I cherry pick `aiding.ts` across?

lenkan: I believe that is the exact same error I reproduced in #252. I created the issue in keria that needs resolution https://github.com/WebOfTrust/keria/issues/230. I have not yet debugged keria for that issue.

iFergal: I made some other adjustments in my latest push - `smids`/`rmids` only changed on group creation, not rotation.

There were also some changes in keripy on not using hard coded salts so our tests can't assume the agent AID anymore.

Once https://github.com/WebOfTrust/keria/pull/243 merges, the only failing test is the one you described for vlei issuance. Haven't got a chance to support on that yet.

PR #239, opened 2024-03-28
fix hex encoded seq num issue when sn is greater than 10

Arsh-Sandhu:
- fix "too large to encode" hex seq num issue when sn is greater than 10
- add singlesig-rot-bulk.test.ts to test sn greater than 10


lenkan:
This was fixed in a conflicting way in development through https://github.com/WebOfTrust/signify-ts/pull/216. This has a nice integration test added though.

PR #240, opened 2024-03-28
add optional nonce fields in issue credential args to multisig private credential issuance

Arsh-Sandhu:
- add attributeNonce, rootNonce fields in IssueCredentialArgs interface  to fix multisig private credential issuance. Members joining create-credential can use these fields to set nonce form `exn` message.


lenkan:
This was fixed in a conflicting way in development from #222

PR #241, opened 2024-03-30
Add feature for renaming and deleting aids

rodolfomiranda:
PR that follows KERIA [PR #218](https://github.com/WebOfTrust/keria/pull/218)

### DIFF
--- a/src/keri/app/aiding.ts
+++ b/src/keri/app/aiding.ts
@@ -120,6 +120,36 @@ export class Identifier {
     }
 
     /**
+     * Get information for a managed identifier
+     * @async
+     * @param {string} name Name or alias of the identifier
+     * @param {string} newName New name or alias of the identifier
+     * @returns {Promise<any>} A promise to the identifier information
+     */
+    async rename(name: string, newName: string): Promise<any> {
+        const path = `/identifiers/${name}`;
+        const data = { name: newName };
+        const method = 'PUT';
+        const res = await this.client.fetch(path, method, data);
+        console.log(res.status);
+        return res.json();
+    }
+
+    /**
+     * Delete a managed identifier
+     * @async
+     * @param {string} name Name or alias of the identifier
+     * @returns {Promise<any>} A promise to the identifier information
+     */
+    async delete(name: string): Promise<any> {
+        const path = `/identifiers/${name}`;
+        const method = 'DELETE';
+        const res = await this.client.fetch(path, method, null);
+        console.log(res.status);
+        return;
+    }
+
+    /**
      * Create a managed identifier
      * @async
      * @param {string} name Name or alias of the identifier
@@ -277,8 +307,8 @@ export class Identifier {
         jsondata[keeper.algo] = keeper.params();
 
         const res = await this.client.fetch(
-            '/identifiers/' + name + '?type=ixn',
-            'PUT',
+            '/identifiers/' + name + '/events',
+            'POST',
             jsondata
         );
         return new EventResult(serder, sigs, res);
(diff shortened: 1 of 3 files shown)


rodolfomiranda:
integrations test will fail until KERIA PR is merged

src/keri/app/aiding.ts
```diff
@@ -119,6 +119,36 @@ export class Identifier {
...
+     * Delete a managed identifier
+     * @async
+     * @param {string} name Name or alias of the identifier
+     * @returns {Promise<any>} A promise to the identifier information
+     */
+    async delete(name: string): Promise<any> {
+        const path = `/identifiers/${name}`;
+        const method = 'DELETE';
+        const res = await this.client.fetch(path, method, null);
+        console.log(res.status);
```
lenkan, line 148: Log statement should be removed.

src/keri/app/aiding.ts
```diff
@@ -119,6 +119,36 @@ export class Identifier {
...
+     * @param {string} name Name or alias of the identifier
+     * @param {string} newName New name or alias of the identifier
+     * @returns {Promise<any>} A promise to the identifier information
+     */
+    async rename(name: string, newName: string): Promise<any> {
+        const path = `/identifiers/${name}`;
+        const data = { name: newName };
+        const method = 'PUT';
+        const res = await this.client.fetch(path, method, data);
+        console.log(res.status);
```
lenkan, line 134: Log statement should be removed.

lenkan (commented):
Nice!

Open question... since we are changing the API anyways. Is there currently any benefit of using the event type as an object field key rather than just relying on the event payload `t` property? I'll illustrate with an example what I mean: 

Currently:

```
POST /identifiers/name/events
{
  "rot": { ...rot payload },
  "sigs": [...sigs]
}

POST /identifiers/name/events
{
  "ixn": { ...ixn payload },
  "sigs": [...sigs]
}
```

Rather than:

```
POST /identifiers/name/events
{
  "event": { t: "rot/ixn", ...rot/ixn payload },
  "sigs": [...sigs]
}
```

The switch on backend would be to check `body.event.t === "ixn" or body.event.t === "rot"`.

lenkan:
@rodolfomiranda There is something wrong after the upgrade to 0.2.0-dev0. I tried to debug the multisig-vlei test and found that after the client of GAR1 resolves the oobi for the multisig QVI aid, they still don't have the QVI aid in the contact list.

To get this merged we need to figure out what the problem is there. Perhaps it would be better to create a separate PR for just the upgrade to 0.2.0-dev0.

lenkan:
@rodolfomiranda Did a little bit more digging, and what we get back from the long running operation for resolving the QVI AID is just ({ oobi: "<oobi url>" }), rather than the expected response of the current state of the contact. This only happens here:

https://github.com/WebOfTrust/keria/blob/7ee5ba84c9aed5d2d21fa48d8241d39b06aadb93/src/keria/core/longrunning.py#L239, which only happens if the `obr.cid` is not in kevers. Any ideas?

rodolfomiranda:
> Any ideas?

I've been debugging a lot and saw the same but also found that the problem starts to happens after a new `ixn` or `rot` event  is applied to the multisig. In the `multisig-vlei-issuanse` test, that happens after the `ixn` for the delegation in GEDA aid; in the `multisig` test that happens after rotating the `multisig` aid. In both cases, if you comment out the event (the delegation in the vlei test or the rotation in the multisig test), the problem not longer appears. If not, seems that the `habs` got corrupted.

I think it's related to the changes that were applied in `keripy` and `keria` to track multisig states, such as the `smids` and `rmids` as states. It works well at inception, but something is wrong after and `ixn` or `rot` event. I'm still troubleshooting...

lenkan:
> > Any ideas?
> 
> I think it's related to the changes that were applied in `keripy` and `keria` to track multisig states, such as the `smids` and `rmids` as states. It works well at inception, but something is wrong after and `ixn` or `rot` event. I'm still troubleshooting...

Cross posting for visibility.

There are probably more issues then. In #252, I managed to reproduce a similar situtation for single sig only, where the problem is that a delegator cannot resolve an oobi of the delegatee.

PR #245, opened 2024-04-11
add integration test matrix for keria version

lenkan:
This allows us to run the integration tests using multiple keria versions to make it more clear which versions this library supports. In the future, we can add some logic to the tests that skips them if they are incompatible.


rodolfomiranda:
That's great!

rodolfomiranda (approved):

pfeairheller (approved):

PR #248, opened 2024-04-11
Merge development to main

lenkan:
I thought we could merge this to main as a rebase or with a merge commit (i.e., not squash), then the history will be identical and retargeting the open PRs will be seamless.


pfeairheller:
Seeing these integration tests break makes me think that a matrix of old KERIA versions probably doesn't make sense.  New development is likely (especially at this stage) for breaking changes against new versions of KERIA.

lenkan:
Hm yeah maybe that was a bit premature. But I think the tests that are failing is due to #244 , at least the from the sample of the logs I checked.

lenkan:
@pfeairheller I added #249 that keeps the matrix, but only uses :latest for now. As you pointed out, currently there will be breaking changes that makes it difficult, but in the future it is easy to add the versions that is supported to the workflow and ensure compatibility.

pfeairheller:
New docker image and PyPi development release pushed at weboftrust/keria:0.2.0-dev0

pfeairheller:
> @pfeairheller I added #249 that keeps the matrix, but only uses :latest for now. As you pointed out, currently there will be breaking changes that makes it difficult, but in the future it is easy to add the versions that is supported to the workflow and ensure compatibility.

I did not update the :latest tag on Docker because I think that should be a working version and not dev.

lenkan:
Right. Then, we can use :latest for now, as specified in #249 (actually, 0.1.2 and 0.1.3 works for now as well). Then because both #241 and #234 requires the dev image, we can update it as part of one of those PRs, otherwise we will have failing tests I think until those are merged.

lenkan:
On second thought - doing this as a PR seems to just add confusion with the Github action runs. At first I thought it would be good for visibility.

PR #249, opened 2024-04-11
update keria test matrix to only use latest for now

lenkan:
As mentioned by Phil in #248, it does not make much sense right now to keep running tests against old versions of keria.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,7 +57,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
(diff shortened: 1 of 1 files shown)


rodolfomiranda:
Agree, but now PR #241 needs [0.2.0-dev0](https://hub.docker.com/layers/weboftrust/keria/0.2.0-dev0/images/sha256-2642a0dcfe18ca1c4462d0af5edac85411b498e186b4570dacd4988ba0f3a50e?context=explore).
Can you add it?

pfeairheller (approved):

PR #250, opened 2024-04-13
Fix docs publishing

lenkan:
Changed the documentation publish workflow to be published from github actions rather than a branch. I find this setup much simpler and we avoid having all the published doc files in the repo. I have tried deploying it to my github pages using the workflow and it works: https://lenkan.dev/signify-ts (just deployed for testing the setup, will remove once the PR is meged). See the workflow run: https://github.com/lenkan/signify-ts/actions/runs/8672410431

We should just change this setting once merged:

image


rodolfomiranda (approved):
approved

pfeairheller (approved):

PR #251, opened 2024-04-13
Add type information to keeping module

lenkan:
This PR adds type information to keeping.ts module.

When working with signify-ts, I find that this module is the source for a lot of missing type information that makes the library difficult to work with, so I tried to add as much as possible to make the library easier to use and safer to refactor.

I could not find a good place to put the interfaces for the "HabState" types, so they are in a file called "state.ts" at the moment. Please give feedback on the solution and type information.


rodolfomiranda:
That's a huge improvement. Thanks!

rodolfomiranda (approved):
LGTM

PR #253, opened 2024-04-26
Fix size from bytes instead of string. Need for serder with special characters

rodolfomiranda:
This PR fix KERIA issue https://github.com/WebOfTrust/keria/issues/201

The `sizeify` method in `Serder` class was computing the size base in the stringifyed json,  but it should be from the bytes, so the version string was created with the wrong size preventing to pass the validation in `keripy`

### DIFF
--- a/src/keri/core/serder.ts
+++ b/src/keri/core/serder.ts
@@ -153,7 +153,7 @@ export function sizeify(
     }
 
     let raw = dumps(ked, kind);
-    const size = raw.length;
+    const size = new TextEncoder().encode(raw).length;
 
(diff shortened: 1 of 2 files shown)


lenkan (commented):
Nice find @rodolfomiranda. I added a test for this and created a PR towards the head of this PR https://github.com/roots-id/signify-ts/pull/1.

It would be nice to add a simple unit test for it, so if you can merge it, copy it, or just create your own test, that would be awesome.

rodolfomiranda:
> It would be nice to add a simple unit test for it

unit test added

lenkan (approved):

PR #254, opened 2024-04-29
fix credentialing field ordering and related singlesig-vlei-issuance test

2byrds:
1. Fixes ordering of u field.
2. Modifies signed fetch to give the caller control over Request related args.
3. Improves reliability of singlesig-vlei-issuer test


src/keri/app/clienting.ts
```diff
@@ -261,33 +266,25 @@ export class SignifyClient {
             new Date().toISOString().replace('Z', '000+00:00')
         );
 
+        if (contentType !== null) {
+            headers.set('Content-Type', contentType);
+        }
+
         if (data !== null) {
-            headers.set('Content-Length', data.length);
+            headers.set('Content-Length', data.length.toString());
```
lenkan, line 274: This was discussed during a PR for the "internal" fetch function. See here: https://github.com/WebOfTrust/signify-ts/pull/233#discussion_r1526156253

- `fetch` will calculate the body length and set this header for you, so you probably do not need to set it unless you have special requirements.
- In any case, data.length does not handle multi byte characters. E.g. `"ö".length === 1` but `new TextEncoder().encode("ö").length === 2`.

src/keri/app/clienting.ts
```diff
@@ -235,15 +235,20 @@ export class SignifyClient {
...
      * @param {string} aidName Name or alias of the AID to be used for signing
      * @returns {Promise<Response>} A promise to the result of the fetch
      */
     async signedFetch(
         url: string,
         path: string,
         method: string,
-        data: any,
+        data: string,
+        contentType: string,
```
lenkan, line 251: Imo, we are starting to get too many positional parameters here. I suggest a breaking change and move the interface towards what `fetch` has. I.e. `fetch(url: string, init: RequestInit)`

src/keri/app/credentialing.ts
```diff
@@ -211,6 +212,7 @@ export class Credentials {
         const [, acdc] = Saider.saidify({
             v: versify(Ident.ACDC, undefined, Serials.JSON, 0),
             d: '',
+            u: privacy ? new Salter({}).qb64 : undefined,
```
lenkan, line 215: Is this because the fields needs to be in a certain order? That is a rather awkward spec requirement, JSON is unordered. Perhaps this requirement should be enforced by KERIA before sending the message on?

See https://www.rfc-editor.org/rfc/rfc7159

> An object is an unordered collection of zero or more name/value pairs, where a name is a string and a value is a string, number, boolean, null, object, or array.

You cannot trust that all signify client implementations are ordering object fields unless they manually construct the JSON. Most JavaScript implementations would do this though.

With that being said, perhaps the most pragmatic approach is to explicitly order _all_ the fields in the ACDC credential, assuming the top level keys would always be in the order [v, d, u, i, ri, s, a, e, r] you can do

```
const [, acdc] = Saider.saidify({
        v: versify(Ident.ACDC, undefined, Serials.JSON, 0),
        d: '',
        u: args.u,
        i: args.i,
        ri: args.ri,
        s: args.s,
        a: subject,
        e: args.e,
        r: args.r,
     })
```

If you also need to allow for additional unknown properties to be added to the end:

```
const { u, i, ri, s, e, r, _a, ...restArgs } = args;
const [, acdc] = Saider.saidify({
        v: versify(Ident.ACDC, undefined, Serials.JSON, 0),
        d: '',
        u: u,
        i: i,
        ri: ri,
        s: s,
        a: subject,
        e: e,
        r: r,
        ...restArgs
})
```

examples/integration-scripts/singlesig-vlei-issuance.test.ts
```diff
@@ -505,18 +505,21 @@ async function getOrIssueCredential(
...
 
-    const issResult = await issuerClient.credentials().issue(issuerAid.name, {
-        ri: issuerRegistry.regk,
-        s: schema,
-        u: privacy ? new Salter({}).qb64 : undefined,
-        a: {
-            i: recipientAid.prefix,
-            u: privacy ? new Salter({}).qb64 : undefined,
-            ...credData,
+    const issResult = await issuerClient.credentials().issue(
```
lenkan, line 508: If my suggestion in the `credentialing` file is handled, we do not need this change.

examples/integration-scripts/vlei-verification.test.ts
lenkan, line 1: Imo, we can probably wait with this type of test. It should at least be its own PR so the discussion can be scoped around the topic.

examples/integration-scripts/singlesig-vlei-issuance.test.ts
```diff
@@ -526,7 +529,7 @@ async function getOrIssueCredential(
     return credential;
 }
 
-async function getGrantedCredential(
+export async function getGrantedCredential(
```
lenkan, line 532: Perhaps this can be moved do a test utils file for multiple tests to consume instead.

examples/integration-scripts/singlesig-vlei-issuance.test.ts
```diff
@@ -21,7 +21,7 @@ const { vleiServerUrl } = resolveEnvironment();
 const QVI_SCHEMA_SAID = 'EBfdlu8R27Fbx-ehrqwImnK-8Cm79sqbAQ4MmvEAYqao';
 const LE_SCHEMA_SAID = 'ENPXp1vQzRF6JwIuS-mp2U8Uf1MoADoP_GqQ62VsDZWY';
 const ECR_AUTH_SCHEMA_SAID = 'EH6ekLjSr8V32WyFbGe1zXjTzFs9PkTYmupJ9H65O14g';
-const ECR_SCHEMA_SAID = 'EEy9PkikFcANV1l7EHukCeXqrzT1hNZjGlUk7wuMO5jw';
+export const ECR_SCHEMA_SAID = 'EEy9PkikFcANV1l7EHukCeXqrzT1hNZjGlUk7wuMO5jw';
```
lenkan, line 24: Perhaps these constants can be moved to a "schemas.ts" type of file so multiple tests can consume them?

package-lock.json
lenkan, line 1: There should not be any changes to this file unless you have installed any new packages, but I cannot see any changes in any package.json.

src/keri/app/credentialing.ts
lenkan: Follow up. Depending on what the KERI/ACDC requirements are for the field ordering. If it is only `[v, d, u]` that are required in that order, you could get away with only picking the `u` field from the second suggestion.

rodolfomiranda: See ACDC field ordering requirements in the following `keripy` code:
https://github.com/WebOfTrust/keripy/blob/fd7e98a26d1e5ce42997a1afc0ee52a62da9a5aa/src/keri/core/serdering.py#L497-L545

lenkan: @rodolfomiranda Thanks for the link. From that it looks like you don't need to support additional properties. Edited my commend.

examples/integration-scripts/singlesig-vlei-issuance.test.ts
2byrds: Based on the other changes I have removed the export.
I have created a new issue for a proper refactor of the reusable test related features that singlesig-vlei-issuance provides https://github.com/WebOfTrust/signify-ts/issues/256

src/keri/app/clienting.ts
2byrds: Removed.

src/keri/app/clienting.ts
2byrds: Aligned parameters to be similar to fetch

src/keri/app/credentialing.ts
2byrds: Used @lenkan suggestion.

package-lock.json
2byrds: Reverted

examples/integration-scripts/singlesig-vlei-issuance.test.ts
2byrds: agree

examples/integration-scripts/singlesig-vlei-issuance.test.ts
```diff
@@ -185,16 +185,18 @@ test('singlesig-vlei-issuance', async function run() {
...
+    let qviRetry = 0;
+    while (!qviCredHolder && qviRetry < 5) {
         await sendGrantMessage(gleifClient, gleifAid, qviAid, qviCred);
         await sendAdmitMessage(qviClient, qviAid, gleifAid);
 
         qviCredHolder = await retry(async () => {
             const cred = await getGrantedCredential(qviClient, qviCred.sad.d);
-            assert(cred !== undefined);
-            return cred;
+            qviRetry += 1;
```
lenkan, line 195: Does the `retry` function not already do retries for you? You can set `maxRetries` in the second parameter.

lenkan (commented):
Cool! Much easier to review now. I don't understand the changes with the retries? Does the `retry` function not do this for you?

Also, if not too much effort, can the changes to signedFetch be separated to it's own PR to make it easier to review the history? I feel these are two unrelated changes.

examples/integration-scripts/singlesig-vlei-issuance.test.ts
2byrds: It does if I activate it with the necessary parameters :) I believe maxRetries is unset by default so I've provided reasonable defaults. And have removed the unnecessary loops.

Issue #227, opened 2024-02-26
Multisig join never resolves


daviddm:
I've made a test where I try to add a third member to a already existing multisig with 2 members.
When the third member tries to join the multisig the join-operation never resolves.

https://github.com/WebOfTrust/signify-ts/commit/4b7bd582572785ca7f073ccab8064051e5b42ef6#diff-4b7bd582572785ca7f073ccab8064051e5b42ef6

2byrds:
From our keri dev meeting, Provenant along with @pfeairheller recently added two people to their 'migrated' wallet (command line to signify/keria). Signifypy has a series of scripts to test this functionality (for the migration). Provenant did use signify-ts for the two new participants. This was a delegated multi-sig that required approval.
Supplying versions is helpful in order to determine where the fix is needed.
Possibly a signify-ts update has the fix. Or an improvement in the steps used.
Join functionality specifically.

Sotatek-HocNguyena:
Hi @daviddm. Were you able to call the `groups().join`? When I called it, it threw me a "500 Internal Server Error" everytime.

daviddm:
I updated the test and made a PR with it https://github.com/WebOfTrust/signify-ts/pull/255

When running the tests the member that tries to join the multisig-group is able to do it and can add the agent end-role after "successful" join. However the join operation still does not resolve.

I'm not sure if something has changed since February, but I know that i didn't try to just disregard the operation status and continue as if it was successful.

So maybe this isn't a signify-ts issue but instead a keria issue?

## Versions
Keria: 0.1.3
Signify-ts: latest-main fddaff2

2byrds:
Thank you for the additions! I'll be taking a look at this next week if it hasn't been addressed by then

lenkan:
I tried fixing that line and run against a local keria image, then it resolves. https://github.com/WebOfTrust/keria/pull/236

lenkan:
This will be resolved once we can merge https://github.com/WebOfTrust/signify-ts/pull/258

2byrds:
@daviddm can this be closed now per https://github.com/WebOfTrust/signify-ts/pull/258 ?

lenkan:
@2byrds Yes, thanks for notifying.

PR #255, opened 2024-04-30
add integration test for aid to join multisig

daviddm:
This PR includes multisig test join and illustrates the problem in https://github.com/WebOfTrust/signify-ts/issues/227.

The integration test creates a multisig-group with 2 members and threshold 1.
Member 1 adds a new member(member3) to the group and does a double rotation for the new member to be able to join and add the end-role by himself.


lenkan:
I added the waitOperation call. Needs a new Keria image with https://github.com/WebOfTrust/keria/pull/236 to make the tests pass.

daviddm:
I made some bad git-stuff.
Remade the PR https://github.com/WebOfTrust/signify-ts/pull/258

PR #259, opened 2024-05-08
signedFetch more generic/configurable aligning it with fetch api

2byrds:


src/keri/app/clienting.ts
```diff
@@ -254,42 +258,26 @@ export class SignifyClient {
             keeper.signers[0].verfer
         );
 
-        const headers = new Headers();
-        headers.set('Signify-Resource', hab.prefix);
+        let headers = req.headers;
```
lenkan, line 261: Should be enough with `const headers = new Headers(req.headers)` I think?

src/keri/app/clienting.ts
```diff
@@ -254,42 +258,26 @@ export class SignifyClient {
...
-        if (data !== null) {
-            headers.set('Content-Length', data.length);
-        } else {
-            headers.set('Content-Length', '0');
-        }
         const signed_headers = authenticator.sign(
-            headers,
-            method,
+            new Headers(headers),
+            headers.get('method')!,
```
lenkan, line 275: Should this not be `req.method` or actually `req.method ?? "GET"` because GET should be the default?

lenkan (commented):
Had a quick look with some initial questions, will try it out later.

src/keri/app/clienting.ts
2byrds: nice! fixed.

src/keri/app/clienting.ts
2byrds: totally should. i updated the clienting.test to check the header signature in hopes of catching mistakes like that.

src/keri/app/clienting.ts
```diff
@@ -232,19 +233,23 @@ export class SignifyClient {
...
+     *              If the data is a CESR JSON string then you should also set contentType to 'application/json+cesr'
+     *              If the data is a FormData object then you should not set the contentType and the browser will set it to 'multipart/form-data'
+     *              If the data is an object then you should use JSON.stringify to convert it to a string and set the contentType to 'application/json'
+     *     - contentType Content type of the request.
      * @returns {Promise<Response>} A promise to the result of the fetch
      */
     async signedFetch(
+        aidName: string,
         url: string,
         path: string,
```
lenkan, line 251:
```suggestion
        url: string,
```

I personally don't see a value in separating these args. In fetch which is a familiar API you do not separate the host and the path. They are both part of the URL.

src/keri/app/clienting.ts
```diff
@@ -232,19 +233,23 @@ export class SignifyClient {
     /**
      * Fetch a resource from from an external URL with headers signed by an AID
      * @async
+     * @param {string} aidName Name or alias of the AID to be used for signing
      * @param {string} url URL of the resource
      * @param {string} path Path to the resource
```
lenkan, line 238:
```suggestion
     * @param {string} url URL of the resource
```

src/keri/app/clienting.ts
lenkan: To get the for the signature you can simply do `const pathname = new URL(url).pathname`. This also strips the query params so the `path.slit["?"]` is not needed on L272

lenkan (changes requested):
Sorry for the delay.

I had some very minor suggestions but I think it looks good otherwise.

src/keri/app/clienting.ts
2byrds: I think whoever defined the method signature was trying to explicitly align with https://datatracker.ietf.org/doc/html/rfc9421#name-path but i'm okay with aligning with fetch and having the method handle the url parts as you specified. Let me know if the RFC alignment alters your thoughts.

Also, browser extensions that use signify-ts should sign headers the same way as signify-ts, but they don't do the fetch. So I'll refactor this method to take a url but not do the fetch. Let me know if you have thoughts on that.

lenkan: Yeah, good idea. I think that makes sense. So you basically return a "signed" RequestInit? Imo, that makes it much more useful.

If you make sure the `createSignedRequest` returns a `Request` (available as a type in the global namespace in both node.js and browsers), the following usage would be supported and be very  straightforward for anyone to use.


```typescript
   // SignifyClient
   async createSignedRequest(aidName: string, url: string, requestInit: RequestInit): Promise<Request> {
   }
  

const request = await client.createSignedRequest("aidname", url, initialRequestInit);
await fetch(request)
```

2byrds: > Yeah, good idea. I think that makes sense. So you basically return a "signed" RequestInit? Imo, that makes it much more useful.
> 
> If you make sure the `createSignedRequest` returns a `Request` (available as a type in the global namespace in both node.js and browsers), the following usage would be supported and be very straightforward for anyone to use.
> 
> ```ts
>    // SignifyClient
>    async createSignedRequest(aidName: string, url: string, requestInit: RequestInit): Promise<Request> {
>    }
>   
> 
> const request = await client.createSignedRequest("aidname", url, initialRequestInit);
> await fetch(request)
> ```

great! working on  that update now

2byrds:
@lenkan latest changes per your recommendations have been added.

examples/integration-scripts/singlesig-vlei-issuance.test.ts
```diff
@@ -108,9 +108,9 @@ const OOR_AUTH_RULES = LE_RULES;
 
 const CRED_RETRY_DEFAULTS = {
     maxSleep: 1000,
-    minSleep: 10,
+    minSleep: 1000,
```
lenkan, line 111: I suppose this change is because with the previous settings, we did maximum 5 retries with a timeout that exponentially increased from 10ms to 1000ms which would yield something that is a lot less than 5 seconds.

A problem with this setting is that the timeout value does not make a difference, since the maxRetries is set to 5 and both minSleep and maxSleep is 1000ms, the retry will timeout at 5*1000ms = 5 seconds

Maybe something like this is better if you need a timeout of 30s (I doubt we need 30s though, maybe 10s is enough)

- minSleep: 100
- maxRetries: undefined
- timeout: 30000

But it is not important, you would only gain maximum one second. I was just curious of this change.

src/keri/app/clienting.ts
```diff
@@ -234,23 +234,21 @@ export class SignifyClient {
      * Fetch a resource from from an external URL with headers signed by an AID
```
lenkan, line 234:
```suggestion
     * Create a Signed Request to fetch a resource from an external URL with headers signed by an AID
```

src/keri/app/clienting.ts
```diff
@@ -234,23 +234,21 @@ export class SignifyClient {
...
+     * @param {string} url URL of the requested resource
      * @param {RequestInit} req Request options should include:
      *     - method: HTTP method
      *     - data Data to be sent in the body of the resource.
      *              If the data is a CESR JSON string then you should also set contentType to 'application/json+cesr'
      *              If the data is a FormData object then you should not set the contentType and the browser will set it to 'multipart/form-data'
      *              If the data is an object then you should use JSON.stringify to convert it to a string and set the contentType to 'application/json'
      *     - contentType Content type of the request.
-     * @returns {Promise<Response>} A promise to the result of the fetch
+     * @returns {Promise<Request>} A promise to the result of the fetch
```
lenkan, line 245:
```suggestion
     * @returns {Promise<Request>} A promise to the created Request
```

lenkan (commented):
Looks really good. Some minor adjustments to the js docs, they were left with the original description so they would be misleading to consumers of the library.

2byrds:
thank you for fixes @lenkan i have updated it per your suggestions

package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "0.2.1",
+  "version": "0.2.2",
```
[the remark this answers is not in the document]
2byrds: I wanted to signal this is a significant change. Happy to handle that however u suggest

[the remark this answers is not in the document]
lenkan: I think the version can be bumped in separate PR. The packages is published manually now anyway and I think the version is bumped using `npm version`. So I vote for leaving this part out of this PR. Could add a "Breaking: " prefix to the PR though.

lenkan (approved):
LGTM

PR #260, opened 2024-05-11
add workflow for publishing npm package

lenkan:
This is a suggestion of an automated workflow for publishing npm packages.

Process for publishing the development tags would be:

1. Create a PR for the next intended npm package version `x.y.z`
   - Most often this would the next minor release for new features. While we are still on v0, we also include breaking changes here. Once we hit v1, we will have a more strict process for how to bump version numbers. Perhaps something like this https://www.npmjs.com/package/semantic-release.
1. Once merged, a maintainer can publish next dev-version Actions => Publish NPM => Run workflow
   - image
1. The workflow does this
   1. Set the version fo `x.y.z-dev.<commit-sha>`
   2. Publish the npm package to npm registry
   3. Creates a git tag `vx.y.z-dev.<commit-sha>`
   4. Push git tag

If this works well, we can trigger this workflow to run on push to main for a completely automated publishing flow.

Once the next release is ready you change the `npm dist tag` input to `latest`. Then the workflow does

1. Publish the npm package with version `x.y.z` to npm registry
2. Creates a git tag `x.y.z`
3. Push git tag

This can then also be easily extended to utilize https://cli.github.com/manual/gh_release_create for automatic github release creation with changelog.

# Usage

If consumers of the library want to install the lastest dev version they would do:

```
npm i --save signify-ts@dev
```

They can also explicitly install the version from a commit using

```
npm i --save signify-ts@vx.y.z-dev.<commit-sha>
```

To install the latest released version they would continue to use

```
npm i --save signify-ts@latest
```

Or simply use a version number

```
npm i --save signify-ts@x.y.z
```

As long as they use a lock file, the version that they have installed in their project will not change unless they explicitly want to upgrade.

# Discussion

In order to make this work, someone with push access to the npm package `signify-ts` needs to generate a publishing token and store in github secrets. These secrets are not accessible from forks.

This would make it easier for consumers of this package to install and test the latest versions of the package without having to reference the repository and pull all devDependencies. It would also make it easier for maintainers to publish the packages.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -1,0 +1,61 @@
+name: Publish NPM
+
+on:
+  workflow_dispatch:
+    inputs:
+      dry-run:
+        type: boolean
+        default: true
+        description: Dry run publish
+      dist-tag:
+        type: choice
+        options:
+          - dev
+          - latest
+        default: dev
+        description: Npm dist tag
+jobs:
+  publish:
+    name: Publish NPM
+    permissions:
+        contents: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@v4
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: '20'
+          cache: 'npm'
+          registry-url: 'https://registry.npmjs.org'
+
+      - name: Install dependencies
+        run: npm ci --ignore-scripts
+
+      - name: Set package name from vars
+        if: ${{ vars.NPM_PACKAGE_NAME }}
+        run: |
+          cat package.json | jq -r '.name = "${{ vars.NPM_PACKAGE_NAME }}"' > package.json.tmp
+          mv package.json.tmp package.json
+
+      - name: Set version from commit
+        if: ${{ inputs.dist-tag == 'dev' }}
+        run: |
+          npm version --no-git-tag-version $(cat package.json | jq .version -r)-dev.$(git rev-parse --short HEAD)
+
+      - name: Create git tag
+        run: git tag $(cat package.json | jq .version -r)
+
+      - name: Publish dev package (Dry run)
+        if: ${{ inputs.dry-run == true }}
+        run: npm publish --tag "${{ inputs.dist-tag }}" --dry-run
+
+      - name: Publish dev package
+        if: ${{ inputs.dry-run == false }}
+        env:
+          NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
+        run: |
+            npm publish --tag "${{ inputs.dist-tag }}"
+            git tag v$(cat package.json | jq .version -r)
+            git push origin  v$(cat package.json | jq .version -r)
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -3,9 +3,11 @@ on:
   push:
     branches:
       - 'main'
+      - 'development'
   pull_request:
     branches:
       - 'main'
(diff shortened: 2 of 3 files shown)


.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,54 @@
...
+        run: |
+          cat package.json | jq -r '.name = "${{ vars.NPM_PACKAGE_NAME }}"' > package.json.tmp
+          mv package.json.tmp package.json
+      - name: Set version from commit
+        if: ${{ inputs.dist-tag == 'dev' }}
+        run: |
+          npm version --no-git-tag-version $(cat package.json | jq .version -r)-dev.$(git rev-parse --short HEAD)
+      - name: Create git tag
+        run: git tag $(cat package.json | jq .version -r)
+      - name: Install dependencies
```
kentbull, line 42: We should add caching here since this will be run frequently, on every push to dev.

.github/workflows/publish-npm.yml
```diff
@@ -0,0 +1,54 @@
...
+        with:
+          node-version: '20'
+          cache: 'npm'
+          registry-url: 'https://registry.npmjs.org'
+      - name: Set package name from vars
+        if: ${{ vars.NPM_PACKAGE_NAME }}
+        run: |
+          cat package.json | jq -r '.name = "${{ vars.NPM_PACKAGE_NAME }}"' > package.json.tmp
+          mv package.json.tmp package.json
+      - name: Set version from commit
```
kentbull, line 36: Is this missing setting a version for "latest" from main?

kentbull (approved):

.github/workflows/publish-npm.yml
lenkan: No, it is only when we release the dev-releases that we need to append the suffix. For example if the version field in the package.json is "0.2.1", we will release it as that, without changes.

.github/workflows/publish-npm.yml
lenkan: This should already be the case since using the cache property of "setup-node" action: https://github.com/actions/setup-node/tree/v4/?tab=readme-ov-file#caching-global-packages-data

I moved the "Install dependencies" step to just after that the "Setup node" to make it clearer.

.github/workflows/publish-npm.yml
lenkan: Another approach that I thought of is to completely omit the "version" field in package.json and just set it from the workflow, or when creating a tag. I think both ways are fine.

lenkan:
Update: I thought a bit more about this and I think the main benefit is to make it easier to publish dev versions automatically. However, I have second thoughts about putting the "logic" for determining the version number in the workflow. It is probably better to but in a script, the workflow would just invoke the script using appropriate environment variables. Can also add a Makefile similar to the other repos.

Essentially I'd like the end result to be something like:

1. Push changes to main
2. A workflow automatically (or after manual approval) publishes the package under x.y.z-dev.<commit-sha> version.
3. Some time passes and a the package is ready for new release
4. Maintainer starts workflow with "latest" dist tag
5. Workflow publishes the package under x.y.z version (from package.json) 
6. Maintainer creates a PR and bumps version to next minor x.w.0
7. Following PRs will be merged and released under x.w.0-dev.<commit-sha> until x.w is ready to be released.

I have not thoroughly considered how to handle release branches.

pfeairheller:
@lenkan The secret NPM_DEPLOY_SECRET has been created with WRITE access to signify-ts.

pfeairheller (approved):

PR #262, opened 2024-05-12
fix: remove development from workflow triggers

lenkan:
Removing the development branch from the workflow triggers as we do not use that branch anymore.


rodolfomiranda (approved):
LGTM

PR #263, opened 2024-05-16
fix: export hab state types from main export

lenkan:
I have noticed it is inconvenient that the types returned from "identifiers().get()" is not exported by the package. This PR fixes that.


rodolfomiranda (approved):
LGTM

PR #264, opened 2024-05-22
fix: delete contact by prefix

iFergal:
The [endpoint on KERIA](https://github.com/WebOfTrust/keria/blob/a1a9b02ceda45fb83885f8ad97cc8de78cb96e45/src/keria/app/aiding.py#L1526) doesn't return a JSON response, so deleting via contact is throwing an error:
```
SyntaxError: Unexpected end of JSON input
  at JSON.parse (<anonymous>)
```

I can add integration tests to cover contacts if we think it's worthwhile but I'd need to find time.


lenkan (approved):
LGTM

PR #265, opened 2024-05-30
Fix non-transferable prefix

rodolfomiranda:
Non transferible AIDs were created with the prefix as the hash of the `icp` instead of the public key itself.


lenkan (approved):
LGTM

PR #268, opened 2024-07-16
Revert "feat: single-sig IPEX apply, offer, agree"

iFergal:
Reverts WebOfTrust/signify-ts#234


lenkan:
@iFergal I think the failing build is due to the update from keria:0.1.2 to keria:0.1.3. `curl` is no longer available in the image, we have to use wget to do the healthcheck. I believe it is safe to merge this and then just open a PR with the new healthcheck.

lenkan (approved):

PR #272, opened 2024-07-30
feat: IPEX apply, offer, agree

iFergal:
Apply, offer, agree IPEX messages. Covers single and multi-sig as multi-sig is just a normal `/ipex/*` message (multi-signed) embedded in `/multisig/exn`.

On the recipient field:
- `rp` (to be renamed to `ri`) is all that's needed in exn messages, so recipient always specified
- no extra harm in including `a.i` - so I haven't removed the code that auto adds this based on the recipient field
  - once keripy has moved to `rp`/`ri` fully we can remove it from Signify.
  - for now, less likely to break things this way.

The `recp` field can still be removed on KERIA but that isn't needed for this PR.

### DIFF
--- a/examples/integration-scripts/credentials.test.ts
+++ b/examples/integration-scripts/credentials.test.ts
@@ -44,6 +44,10 @@ let holderAid: Aid;
 let verifierAid: Aid;
 let legalEntityAid: Aid;
 
+let applySaid: string;
+let offerSaid: string;
+let agreeSaid: string;
+
 beforeAll(async () => {
     [issuerClient, holderClient, verifierClient, legalEntityClient] =
         await getOrCreateClients(4);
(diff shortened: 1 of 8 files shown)


lenkan (commented):
LGTM. One thing to consider in the IpexXXXArgs interfaces is if we should be more descriptive in the property names for grant/apply/offer etc... So:

```
interface IpexAdmitArgs {
   grantSaid: string
]
```

What do you reckon?

iFergal:
thanks, agreed @lenkan I also changed it to `schemaSaid` since it only works with SAIDs right now

iFergal:
@lenkan can this be merged, or awaiting review from @rodolfomiranda? I couldn't make the dev call

lenkan:
I think we can merge it now. No one seems to have any objections.

PR #273, opened 2024-08-27
fix: send exn to actual recipient only

iFergal:
exn messages have a specific recipient (`rp` field).

Right now for multi-sig:
- creates first exn message with `rp` field of first member
- this message is sent to all members
- returns early (bug in itself)

This change will just send the specific message to the specific member, and not return early.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7572,12 +7572,12 @@
       ]
     },
     "node_modules/micromatch": {
-      "version": "4.0.5",
-      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.5.tgz",
-      "integrity": "sha512-DMy+ERcEW2q8Z2Po+WNXuw3c5YaUSFjAO5GsJqfEl7UjvtIuFKO6ZrKvcItdy98dwFI2N1tg3zNIdKaQT+aNdA==",
+      "version": "4.0.8",
+      "resolved": "https://registry.npmjs.org/micromatch/-/micromatch-4.0.8.tgz",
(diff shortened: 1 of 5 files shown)


iFergal:
ok seems to have broken 1 test, interesting.. checking

package-lock.json
iFergal, line 1: Changes here for audit check on pipeline

src/keri/app/exchanging.ts
```diff
@@ -78,24 +78,19 @@ export class Exchanges {
         payload: Dict<any>,
         embeds: Dict<any>,
         recipients: string[]
-    ): Promise<any> {
-        for (const recipient of recipients) {
+    ): Promise<any[]> {
```
iFergal, line 81: we should get proper types here instead of `any` - don't have capacity right now but will try to sweep things in coming weeks

lenkan: Hm, this actually now returns an array of `Promise`. So more correct would be `Promise<any>[]`.

However, a better option could be to change this method to have _one_ recipient and adjust from the call site accordingly.

iFergal: good point, it was right and i incorrectly refactored and TS didnt pick it up due to any - will fix.

But not so sure on changing the method to only have one recipient as this brings a lot of for loops at the call sites, which will make every wallet/app using multi-sig need to refactor everywhere.

iFergal: and i broke tests again lol

PR #282, opened 2024-10-11
feat: get credential registry state (tever vcstate)

iFergal:
This corresponds to WebOfTrust/keria#304 - see description there for more details.

Essentially can get credential state (iss/rev) without needing to already hold the credential (or even be the holder).


src/keri/app/credentialing.ts
```diff
@@ -229,6 +229,23 @@ export interface IpexAdmitArgs {
...
+    i: string,
+    s: string,
+    d: string,
+    ri: string,
+    a: { s: number, d: string },
+    dt: string,
+    et: string
+} & ({
+    et: "iss" | "rev",
+    ra: Record<string, never>
```
iFergal, line 243: This corresponds to `{}` - it's always empty for iss/rev.

src/keri/app/credentialing.ts
```diff
@@ -229,6 +229,23 @@ export interface IpexAdmitArgs {
...
+    vn: [number, number],
+    i: string,
+    s: string,
+    d: string,
+    ri: string,
+    a: { s: number, d: string },
+    dt: string,
+    et: string
+} & ({
+    et: "iss" | "rev",
```
iFergal, line 242: @lenkan `Ilks` is an object, and many other things in that file. Maybe a string enum would be better for our typing?

src/keri/app/credentialing.ts
```diff
@@ -229,6 +229,23 @@ export interface IpexAdmitArgs {
     datetime?: string;
 }
 
+export type CredentialState = {
+    vn: [number, number],
+    i: string,
+    s: string,
+    d: string,
+    ri: string,
+    a: { s: number, d: string },
```
iFergal, line 238: (really is an int, unlike most `s` fields)

examples/integration-scripts/credentials.test.ts
```diff
@@ -248,6 +248,15 @@ test('single signature credentials', async () => {
         await waitOperation(issuerClient, op);
     });
 
+    await step('holder can get the credential status before or without holding', async () => {
```
iFergal, line 251: If I can get transaction state noticing working on keripy, this could be extended nicely to check for revocations w/o IPEX.

iFergal:
Ah @2byrds forgot we need a new dev release of KERIA. Do you have permissions for that?

2byrds:
> Ah @2byrds forgot we need a new dev release of KERIA. Do you have permissions for that?

lets see @iFergal ! I have launched the docker dev release (same version 0.2.0-dev3) https://github.com/WebOfTrust/keria/actions/runs/11315384968

examples/integration-scripts/credentials.test.ts
```diff
@@ -248,6 +248,18 @@ test('single signature credentials', async () => {
         await waitOperation(issuerClient, op);
     });
 
+    await step(
+        'holder can get the credential status before or without holding',
```
lenkan, line 252: For my understanding, is the `/ipex/grant` prior to this required for the holder to be able to query this?

src/keri/app/credentialing.ts
lenkan: @iFergal but it is a sequence number anyway right? If so, Is it the sequence number of the credential state changes, or of the registry event? If the registry event, is it possible we have a similar issue as we had for KEL events where it would not handle `s >= 10 (decimal)` correctly?

examples/integration-scripts/credentials.test.ts
iFergal: Currently, yes. When the discloser sends an `/ipex/grant` it will also send all of the TEL and anchor events in the background.

You can query the TEL too - I tried to fix this in WebOfTrust/keripy#868 - still waiting for feedback though.

src/keri/app/credentialing.ts
iFergal: Yes, it's still a sequence number. It's the sn of the anchoring event in the KEL for the current credential event. `sn` is used from below by keripy when exporting this state.

A mix of int and hex seems to be used throughout keripy, I believe it just needs to be hex when it's actually part of a KERI event and this is just an export of current state.

I think it's OK for Signify to consider `s: number` as a int, and if it's actually a hex being thrown out cast as number then it's a bug?

```
    @property
    def sn(self):
        """
        Property sn: sequence number as int
        Returns .raw converted to int
        """
        return int.from_bytes(self.raw, 'big')

    @property
    def snh(self):
        """
        Property snh:  sequence number as hex
        Returns .sn int converted to hex str
        """
        return f"{self.sn:x}"  # "{:x}".format(self.sn)
```

iFergal:
thanks @2byrds - hard to tell if the issues are from the changes I made to the Makefile or were already broken, as it seems to be the only time it was ran from the pipeline: https://github.com/WebOfTrust/keria/actions/workflows/publish-keria.yml

test/app/credentialing.test.ts
```diff
@@ -315,6 +315,17 @@ describe('Credentialing', () => {
...
+        await credentials.state(mockCredential.sad.ri, mockCredential.sad.d);
+        lastCall = fetchMock.mock.calls[fetchMock.mock.calls.length - 1]!;
+        assert.equal(
+            lastCall[0]!,
+            url +
+                '/registries/EGK216v1yguLfex4YRFnG7k1sXRjh3OKY7QqzdKsx7df/EMwcsEMUEruPXVwPCW7zmqmN8m0I3CihxolBm-RDrsJo'
+        );
+        assert.equal(lastCall[1]!.method, 'GET');
+        assert.equal(lastCall[1]!.body, null);
+        console.log(JSON.stringify(lastCall, null, 2));
```
lenkan, line 328:
```suggestion
```

lenkan: To avoid cluttering the log output when running tests

lenkan (commented):
LGTM, can we just remove the console.log from the unit test?

test/app/credentialing.test.ts
iFergal: done, agree, missed that!

lenkan (approved):

PR #283, opened 2024-10-15
feat: rename managed identifier

iFergal:
I believe this was only added to signifypy following WebOfTrust/keria#218.

### DIFF
--- a/examples/integration-scripts/salty.test.ts
+++ b/examples/integration-scripts/salty.test.ts
@@ -168,5 +168,14 @@ test('salty', async () => {
 
     await assertOperations(client1);
 
+    aid = await client1.identifiers().rename('aid3', 'aid4');
+    assert.equal(aid.name, 'aid4');
+    aid = await client1.identifiers().get('aid4');
+    assert.equal(aid.name, 'aid4');
+    aids = await client1.identifiers().list(2, 2);
+    assert.equal(aids.aids.length, 1);
+    aid = aids.aids[0];
+    assert.equal(aid.name, 'aid4');
+
     console.log('Salty test passed');
 }, 30000);
--- a/src/keri/app/aiding.ts
+++ b/src/keri/app/aiding.ts
@@ -7,7 +7,6 @@ import { MtrDex } from '../core/matter';
 import { Serder } from '../core/serder';
 import { parseRangeHeaders } from '../core/httping';
 import { KeyManager } from '../core/keeping';
-import { Operation } from './coring';
 import { HabState } from '../core/state';
 
 /** Arguments required to create an identfier */
@@ -111,14 +110,31 @@ export class Identifier {
      * Get information for a managed identifier
      * @async
      * @param {string} name Name or alias of the identifier
-     * @returns {Promise<any>} A promise to the identifier information
+     * @returns {Promise<HabState>} A promise to the identifier information
      */
     async get(name: string): Promise<HabState> {
         const path = `/identifiers/${encodeURIComponent(name)}`;
         const data = null;
         const method = 'GET';
         const res = await this.client.fetch(path, method, data);
-        return await res.json();
+        return res.json();
+    }
+
+    /**
+     * Rename managed identifier
+     * @async
+     * @param {string} name Name or alias of the identifier
+     * @param {string} newName New name for the identifier
+     * @returns {Promise<HabState>} A promise to the identifier information after updating
+     */
+    async rename(name: string, newName: string): Promise<HabState> {
+        const path = `/identifiers/${name}`;
+        const method = 'PUT';
+        const data = {
+            name: newName,
+        };
+        const res = await this.client.fetch(path, method, data);
+        return res.json();
     }
 
     /**
@@ -472,7 +488,7 @@ export class Identifier {
             'GET',
             undefined
         );
-        return await res.json();
+        return res.json();
     }
 }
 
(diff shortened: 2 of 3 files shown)


src/keri/app/aiding.ts
```diff
@@ -472,7 +488,7 @@ export class Identifier {
             'GET',
             undefined
         );
-        return await res.json();
+        return res.json();
```
lenkan, line 491: In general, I would say `await res.json()` is preferred. It should keep the stack trace in the event more "intact". Consider what would happen if the response did not have a JSON-parseable response body.

For example:

```
async function getData() {
    const response = await fetch("https://example.com")
    return await response.json()
}


await getData()

// Will produce:
<!doctype html>
^

SyntaxError: Unexpected token '<', "<!doctype "... is not valid JSON
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:5489:19)
    at successSteps (node:internal/deps/undici/undici:5470:27)
    at fullyReadBody (node:internal/deps/undici/undici:4381:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async consumeBody (node:internal/deps/undici/undici:5479:7)
    at async getData (file:///home/lenkan/code/signify-ts/fetch.js:3:12)
    at async file:///home/lenkan/code/signify-ts/fetch.js:7:1

Node.js v20.18.0
```

Where as 

```
async function getData() {
    const response = await fetch("https://example.com")
    return response.json()
}


await getData()

// Would produce

<!doctype html>
^

SyntaxError: Unexpected token '<', "<!doctype "... is not valid JSON
    at JSON.parse (<anonymous>)
    at parseJSONFromBytes (node:internal/deps/undici/undici:5489:19)
    at successSteps (node:internal/deps/undici/undici:5470:27)
    at fullyReadBody (node:internal/deps/undici/undici:4381:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async consumeBody (node:internal/deps/undici/undici:5479:7)
    at async file:///home/lenkan/code/signify-ts/fetch.js:7:1

Node.js v20.18.0
```

In the first example, the asynchronous task of reading the response body is resolved and then parsed within the `getData` function call. In the second example, it is resolved and then parsed in the module scope. So the error stack trace is captured outside of the function body.

You can also read more here about the (deprecated) eslint rule https://eslint.org/docs/latest/rules/no-return-await

examples/integration-scripts/salty.test.ts
```diff
@@ -168,5 +168,14 @@ test('salty', async () => {
 
     await assertOperations(client1);
 
+    aid = await client1.identifiers().rename('aid3', 'aid4');
```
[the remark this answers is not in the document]
lenkan: I am OK with either. Sounds like a good idea though.

lenkan (commented):
Great to finally get this in the API. I had a thought about the use of name vs prefix as the "preferred" way of doing this operation.

I'd rather see `await client1.identifiers().rename(aid3.prefix, 'aid4')` as the preferred method, i.e:

```
PUT /identifiers/prefix
{ "name": "name2" }
```

I find the semantics of

```
PUT /identifiers/name1
{ "name": "name2" }
```

To create a resource at `/identifiers/name2` a little unintuitive.

src/keri/app/aiding.ts
```diff
@@ -111,14 +110,31 @@ export class Identifier {
...
+    }
+
+    /**
+     * Rename managed identifier
+     * @async
+     * @param {string} name Name or alias of the identifier
+     * @param {string} newName New name for the identifier
+     * @returns {Promise<HabState>} A promise to the identifier information after updating
+     */
+    async rename(name: string, newName: string): Promise<HabState> {
```
lenkan, line 130: Have we considered an API that is

```
interface IdentifierInfo {
    name: string
}

async update(prefix: string, info: IdentifierInfo): Promise<HabState> {

}
```

To open up for additional metadata to attach to an identifier that we can update in the future?

src/keri/app/aiding.ts
iFergal: Interesting, thanks for the info! I wasn't aware.

Will fix (and adapt my repos going forward too!)

examples/integration-scripts/salty.test.ts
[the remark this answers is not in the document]
iFergal: OK, I'll leave it for now then as I'm pressed for time.

src/keri/app/aiding.ts
iFergal: For now, I have kept it as `update(name: string, info: IdentifierInfo): Promise<HabState>` to be consistent. If we decide to move to just using a prefix, I'd suggest we open a ticket on KERIA and clean it up there too

lenkan: > Good idea. I'd like to be able to attach additional metadata in the future too.
> 
> Regarding the interface `prefix` vs `name`. This endpoint supports updating by prefix too. I'm changing my entire usage of every API to use the prefix as it makes more sense. Perhaps we should move to only supporting the prefix? It wouldn't be that hard to update the tests, and it'd just be cleaner.

I agree that it makes more sense to only support the prefix. I am all for opening a new issue to track the progress for that. The issues in keria that covered this were initially just about adding support for prefix, so they are closed now. But I believe they both hinted that we should move to only using the prefix eventually:

 - https://github.com/WebOfTrust/keria/issues/217
 - https://github.com/WebOfTrust/keria/issues/202

lenkan (approved):

lenkan:
I will merge this one now since it seems like no one has any objections to the API.

PR #286, opened 2024-10-16
fix: a way to prevent duplicated anchor events for multisig issuance

lenkan:
This PR solves an issue that occurs when a multisig member is trying to join an issuance event after it has been fully signed. They need to do this to be able to import the credential in their database.

This is an alternative approach to the PR by @rodolfomiranda https://github.com/WebOfTrust/signify-ts/pull/271.

I will do the same for `credentials().revoke()` and `registries().create()` if you agree with this approach.

With these changes, I am able to complete this test case: https://github.com/nordlei/vlei-sandbox/blob/main/src/issues/multisig-catchup-problem.test.ts, so we should be able to resolve this issue: https://github.com/WebOfTrust/keria/issues/283. The catch up mechanism would be to simply join all notifications for the multisig group.


src/keri/app/credentialing.ts
```diff
@@ -306,55 +330,65 @@ export class Credentials {
...
-                },
-            ],
-            dig: hab.state.d,
-            version: undefined,
-            kind: undefined,
-        });
+        const sn = new CesrNumber({}, parseInt(hab.state.s, 16) + 1);
+        const anc = new Serder(
+            args.anc ??
+                Saider.saidify({
```
lenkan, line 369: Note: the reason I inlined this is because I think it makes it much more obvious what we are actually posting down to KERIA. I do not think the indirection of calling `interact` provides any readability improvements. But I am OK with either. Just provide your feedback.

iFergal:
@lenkan For IPEX the advantage of how it is set up is that you can also send `/multisig/exn` with an embedded `/ipex/apply` using `submitApply`. KERIA will just handle them differently based on the route.

I kind of like that approach as it's a bit more flexible in general and quite clear from the business logic level: a clean function to create a message or event, and the API submission. This would be my preference in general.

But your change in this PR also makes sense to me if that's preferred by people. I don't like jumping too many levels deep in function calls as it's easy to get lost.

src/keri/app/credentialing.ts
2byrds: Good with me

2byrds:
> @lenkan For IPEX the advantage of how it is set up is that you can also send `/multisig/exn` with an embedded `/ipex/apply` using `submitApply`. KERIA will just handle them differently based on the route.
> 
> I kind of like that approach as it's a bit more flexible in general and quite clear from the business logic level: a clean function to create a message or event, and the API submission. This would be my preference in general.
> 
> But your change in this PR also makes sense to me if that's preferred by people. I don't like jumping too many levels deep in function calls as it's easy to get lost.

@lenkan what do you think about advantages that @iFergal lists here?

iFergal:
I discussed this offline with @lenkan a while ago. It is a bit curious that there's "2 ways" to do multi-sig IPEX which is probably unnecessary complexity. The benefit is less network calls though.

Though based on the eventual outcome of WebOfTrust/keripy#884 it might need to have the calls split anyway.

lenkan:
> @lenkan I updated my signify-ts to this branch. And then tried to modify vlei-sandbox to use the new IssueCredentialArgs interface. But perhaps you already did this and can share? if i can prove that your catchup test works, i can include it in my rollback to prove that member3 is operational in the group again

Hm, when I tried that, I only used npm link locally, so I never had that branch of vlei-sandbox pushed. Let me know if it is still an issue, then I could probably push a branch to vlei-sandbox that uses this for joining the credential.

PR #289, opened 2024-10-18
ESSR support

jimcase:


kentbull:
@jimcase did you decide against adding this to SignifyTS or are you adding this work elsewhere?

PR #290, opened 2024-11-08
feat: agent config interface

iFergal:
To accompany WebOfTrust/keria#318.

Just putting to for review now as I'm just about to go on holidays for a week but when back if we want to merge we'll need a new dev release on KERIA, which I can give a go at doing now.


lenkan (approved):
LGTM

PR #291, opened 2024-11-16
Update README.md with simple example and ready()

daidoji:
Added simple code example to get started and a  note about calling the ready() function for the library to be usable.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,20 @@ Typescript source files needs to be transpiled before running scripts or integra
     npm run build
     ```
 
+-   ready() must be called before library is useable.  Example minimum viable client code.
+    ```javascript
+    import { ready, SignifyClient, Tier } from 'signify-ts';
+
+    await ready();
+    
+    const bran = '0'.repeat(21);
+    const url = 'http://127.0.0.1:3901';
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -40,10 +40,31 @@ The code is built using Typescript and running code locally requires a Mac or Li
...
     ```
 
+-   ready() must be called before library is useable. Example minimum viable client code.
+
+    ```javascript
+    import { ready, SignifyClient, Tier } from 'signify-ts';
+
+    await ready();
+
+    const bran = '0'.repeat(21);
```
iFergal, line 55: We should probably use `randomPasscode` here

daidoji: sure.

iFergal: Yeah since Signify is stateless but it's never shared with KERIA. It's just used to derive the keys locally, and those public keys/events etc are shared with KERIA.

daidoji: @iFergal yeah you're right, I'll remove the comment.

daidoji:
Added #292 to fix failing test action in this PR.

lenkan (approved):

PR #293, opened 2024-11-19
chore: simplify docker compose file

lenkan:
- Use `--wait` option instead of the `deps` service: https://docs.docker.com/reference/cli/docker/compose/up/
- Use yaml anchors to avoid repetitive healthcheck and environment options.


kentbull (approved):
Looks good to me. This is the first I had seen the `--wait` argument. It looks useful.

PR #297, opened 2024-12-12
feat: delete credential by said

iFergal:
Removes credential from DB. KERIA does not yet handle what happens when deleting a chained credential, so maybe beware. I only realised this because I was trying to delete the QVI credential in the integration test and was getting a 404.

Once https://github.com/WebOfTrust/keria/issues/339 addressed, if the result is deleting the chained credentials below that point, the tests here could be updated.

### DIFF
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -21,7 +21,7 @@ services:
             - 7723:7723
 
     keria:
-        image: ${KERIA_IMAGE:-weboftrust/keria}:${KERIA_IMAGE_TAG:-0.2.0-dev5}
+        image: ${KERIA_IMAGE:-weboftrust/keria}:${KERIA_IMAGE_TAG:-0.2.0-dev6}
         environment:
             KERI_AGENT_CORS: 1
             <<: *python-env
--- a/docker-compose.yaml
+++ b/docker-compose.yaml
@@ -27,7 +27,7 @@ services:
             <<: *python-env
         volumes:
             - ./config/keria.json:/keria/config/keri/cf/keria.json
-        command: --config-dir /keria/config --config-file keria --name agent
+        command: start --config-dir /keria/config --config-file keria --name agent
         healthcheck:
             test: wget --spider http://keria:3902/spec.yaml
             <<: *healthcheck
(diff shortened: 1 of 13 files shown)


src/keri/app/notifying.ts
```diff
@@ -59,11 +59,11 @@ export class Notifications {
...
      * @async
      * @param {string} said SAID of the notification
-     * @returns {Promise<any>} A promise to the result of the deletion
+     * @returns {Promise<void>}
      */
     async delete(said: string): Promise<void> {
         const path = `/notifications/` + said;
         const method = 'DELETE';
-        await this.client.fetch(path, method, null);
+        await this.client.fetch(path, method, undefined);
```
iFergal, line 67: `JSON.stringify(undefined)` is `undefined` whereas `JSON.stringify(null)` is `"null"`

docker-compose.yaml
```diff
@@ -21,13 +21,13 @@ services:
...
     keria:
-        image: ${KERIA_IMAGE:-weboftrust/keria}:${KERIA_IMAGE_TAG:-0.2.0-dev5}
+        image: ${KERIA_IMAGE:-weboftrust/keria}:${KERIA_IMAGE_TAG:-0.2.0-dev6}
         environment:
             KERI_AGENT_CORS: 1
             <<: *python-env
         volumes:
             - ./config/keria.json:/keria/config/keri/cf/keria.json
-        command: --config-dir /keria/config --config-file keria --name agent
+        command: start --config-dir /keria/config --config-file keria --name agent
```
iFergal, line 30: @lenkan I know your PR will supersede this, was just putting the minimal changes for this PR.

examples/integration-scripts/credentials.test.ts
```diff
@@ -616,4 +616,18 @@ test('single signature credentials', async () => {
 
         assert.equal(issuerCredential.status.s, '1');
     });
+
+    await step('Holder deletes LE credential', async () => {
+        await holderClient.credentials().delete(leCredentialId);
+        assert.rejects(
```
lenkan, line 622: Never used `assert.rejects` but to me it looks like this would give false positives since this is not awaited.

iFergal: It actually was always checking for me until I put the right string because I guess the next `await` below it takes longer to fetch that this on average. But yes, awaited now! Thanks for spotting.

I also added awaits elsewhere for rejects.

lenkan (approved):

iFergal:
@lenkan good to merge?

PR #298, opened 2024-12-18
chore: enrich package.json

kentbull:
Bump version to 0.3.0-dev0 (latest code from today's `main` branch) and enrich the package.json with useful package metadata to support discovery via NPM.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,17 @@
 {
+  "name": "signify-ts",
   "version": "0.3.0",
+  "description": "Signing at the edge for KERI, ACDC, and KERIA",
+  "keywords": ["keri","acdc","keria","signify","signify-ts","decentralized identity","authentic data","zero trust architecture"],
+  "author": "Phil Feairheller",
+  "homepage": "https://github.com/WebOfTrust/signify-ts",
+  "repo": {
+    "type": "git",
+    "url": "git+https://github.com/WebOfTrust/signify-ts.git"
+  },
+  "bugs": {
+    "url": "https://github.com/WebOfTrust/signify-ts/issues"
+  },
   "license": "Apache-2.0",
   "exports": {
     ".": {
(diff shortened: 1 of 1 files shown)


2byrds (approved):
lgtm

lenkan (changes requested):
If we want to do this, can you run `npm install` locally on your machine to ensure package-lock.json is updated with the new version?

Otherwise, the idea was that the  version field on the `main` branch is the _next_ minor/major version. Then you publish pre-releases from that by temporarily doing an in place edit in package.json before npm publish. That way, you can implement a workflow that publishes a pre-release on every commit to main, and every git tag. To elaborate:

- `version: 0.3.0` on package.json on main.
- Commit with sha `abc123` is pushed to main.
- Workflow triggers and publishes `0.3.0-abc123` to npm
- Commit with sha `zyx123` is pushed to main.
- Workflow triggers and publishes `0.3.0-xyz123` to npm
- Repeat n number of times
- Maintainer decides to publish a beta tag:
   - Create git tag `0.3.0-dev.1`
   - This triggers the workflow that publishes `0.3.0-dev.1` to npm
- Maintainer decides to publish 0.3.0:
   - Create git tag `0.3.0`
   - This triggers the workflow that publishes `0.3.0` to npm
   - Create new commit on main that changes `version: 0.4.0`.
   - Any bug fixes to `0.3.x` can be branched of from the `0.3.0` tag.

lenkan:
I did some work on this before: https://github.com/WebOfTrust/signify-ts/pull/260#issuecomment-2116965535

However, the current `NPM_DEPLOY_SECRET`  does not have sufficient rights to publish. I also have a simplified version of the workflow that allows you to do the same thing locally, but I never finished that work. If we can get a new working deploy secret, either for `signify-ts` or `@weboftrust/signify` package, I can pick it up again over the weekend.

lenkan (approved):

PR #299, opened 2024-12-30
fix: EventResult should await OK response

iFergal:
In places, we were still passing a `Promise<Response>` to `EventResult`. This is problematic because if a non 2xx code is returned as the fetch call will reject after the function has returned.

### DIFF
--- a/src/keri/app/aiding.ts
+++ b/src/keri/app/aiding.ts
@@ -266,7 +266,7 @@ export class Identifier {
         jsondata[algo] = keeper.params();
 
         this.client.pidx = this.client.pidx + 1;
-        const res = this.client.fetch('/identifiers', 'POST', jsondata);
+        const res = await this.client.fetch('/identifiers', 'POST', jsondata);
         return new EventResult(serder, sigs, res);
     }
 
@@ -447,7 +447,7 @@ export class Identifier {
             sigs: sigs,
         };
 
-        const res = this.client.fetch(
+        const res = await this.client.fetch(
             '/identifiers/' + name + '/endroles',
             'POST',
             jsondata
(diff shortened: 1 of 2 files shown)


lenkan (approved):

lenkan:
LGTM, although, it would be nice to implement a test case that verifies this. For example by setting up a unit test with a fetch call that will return a rejected promise.

```
// aiding.test.ts

    it('Should throw error if fetch call fails', async () => {
        const error = new Error(`Fail ${randomUUID()}`);
        client.fetch.mockRejectedValue(error);
        await expect(
            client
                .identifiers()
                .create('aid1', { bran: '0123456789abcdefghijk' })
        ).rejects.toThrow(error);
    });
    
 ```
 
 Currently, it believe the library will NOT throw for non-OK responses. This test will not pass:
 
 ```typescript
    it('Should throw HTTP if fetch resolves with non-JSON', async () => {
        client.fetch.mockResolvedValue(
            new Response('Not JSON', { status: 500 })
        );

        await expect(
            client
                .identifiers()
                .create('aid1', { bran: '0123456789abcdefghijk' })
        ).rejects.toThrow('Fail');
    });
```

This is because the library does not check the HTTP status code. It will not fail until we try to read the message body as JSON, which we do in `EventResult.op()` method.

iFergal:
@lenkan I'd argue this is now covered by the typing. But sure, I can add a test in case a regression happens on the types.

iFergal:
@lenkan I didn't quite get your example. The call _will_ throw an error if the response is e.g. 400 before checking `EventResult.op()`.

https://github.com/WebOfTrust/signify-ts/blob/main/src/keri/app/clienting.ts#L206-L215
```
        const res = await fetch(this.url + path, {
            method: method,
            body: _body,
            headers: final_headers,
        });
        if (!res.ok) {
            const error = await res.text();
            const message = `HTTP ${method} ${path} - ${res.status} ${res.statusText} - ${error}`;
            throw new Error(message);
        }
```

lenkan:
> @lenkan I didn't quite get your example. The call _will_ throw an error if the response is e.g. 400 before checking `EventResult.op()`.
> 
> https://github.com/WebOfTrust/signify-ts/blob/main/src/keri/app/clienting.ts#L206-L215

Ah yes, sorry. I saw that and was just about to write a comment about this. I was mistaken. I thought it didn't, but that is because the `SignifyClient.fetch` is mocked in the test.

lenkan:
Let me know if you want to add anything more, otherwise I can merge it.

iFergal:
@lenkan Thanks, I added the first test for completeness. Nothing else to add now from my side.

PR #304, opened 2025-01-17
feat: ESSR tunneling with KERIA API

iFergal:
This will require a new dev release after https://github.com/WebOfTrust/keria/pull/351 is merged, but locally all the integration tests pass.

HTTP requests between Signify and KERIA are now wrapped/tunneled in ESSR (encrypted HTTP request inside the body of a signed wrapper HTTP request) for confidentiality and general improvements.

`createSignedRequest` can still be used to use the old method of signed headers in case any projects or extensions are using that for web interactions.


edeykholt:
@iFergal Great to see this security hardening!

Since Authenticater / Authenticator is exported and its spelling changed, this might suggest a version change?

Note that WebCrypto in Chrome browser will soon support X25519, so when that's available, it might be a more performant alternative to libsodium when signify-ts is used in browser extensions. Could be a separate backlog item to abstract the crypto library interface.
[[X25519 algorithm of the Web Cryptography API](https://chromestatus.com/feature/6291245926973440)](https://chromestatus.com/feature/6291245926973440)

iFergal:
@edeykholt The `0.3.0` branch was cut recently, so this will be on the next version anyway.

Regarding web crypto, this is great to see but we need more than just X25519 from libsodium. crypto box seal from libsodium provides the hybrid public key encryption described in the SPAC paper.

PR #307, opened 2025-01-27
refactor: align object names with KERIpy and add function docs

kentbull:
This is a work in progress PR. Status will be updated to "Ready for Review" at the end of this week.

## Goals of PR

- Reduce cognitive overhead in comparing SignifyTS code to KERIpy code by aligning class, interface, function, and argument names with KERIpy.
- Add documentation to everything involved in Signify key management.
- Small refactorings to leverage type safety of Typescript's static typing where there is low hanging fruit for easy wins.

## Overview 

This PR does three main things:
- Aligns names with KERIpy where possible or practical.
- Adds function, class, and method documentation in many places.
- Renames things like Keeper, SaltyKeeper, and RandyKeeper to names that more closely match the KERIpy implementation. 

### Example - Keeper to IdentifierManager

As an example, key management in KERIpy occurs in `keeping.Manager`, not `keeping.Keeper`, so the SignifyTS `Keeper` interface, which mimicked the `keeping.Manager` class in KERIpy, was renamed to be `IdentifierManager`. The `keeping.Keeper` class in KERIpy was used to store keys in the LMDB database whereas the Manager was used for creating keypairs, performing inception, rotation, and signing, among other responsibilities not ported to SignifyTS from KERIpy.

The name `Manager` is already taken in SignifyTS by the `manager.Manager` class, yet the responsibilities of the `keeping.Manager` class were being split across both the `manager.Manager` class and the `keeping.Keeper` class. So the existing `keeping.Keeper` class in SignifyTS was renamed to `keeping.IdentifierManager` since that seemed to be a descriptive, clear name describing the fact that implementations of this object perform key pair creation, inception, rotation, and signing.

### Motivation

While reading the Signify source code recently to gain a deeper understanding of key management in Signify I have noticed again a number of naming differences between SignifyTS and KERIpy that are unnecessary and introduce questions of meaning and compatibility. This PR makes the code clear by aligning the codebase with the naming used in the KERIpy codebase.

The significant work of Phil and Kevin, and those who came after them, to port the style, names, and ideas from KERIpy is a strong influence on the motivation to continue that alignment and further it.

### Future work to complete this PR

My current task is to understand the ways stems for Salty key generation and lookup work. I plan to document everything I read as I gain a complete understanding of that code.

### Looking for comments

Please feel free to opine and share your thoughts on what naming policies or related ideas should make it into this PR.

### DIFF
--- a/src/keri/core/keeping.ts
+++ b/src/keri/core/keeping.ts
@@ -11,10 +11,10 @@ import { Prefixer } from './prefixer';
 import { Signer } from './signer';
 import {
     ExternState,
-    GroupState,
+    GroupKeyState,
     HabState,
-    RandyState,
-    SaltyState,
+    RandyKeyState,
+    SaltyKeyState,
     KeyState,
 } from './keyState';
 
@@ -195,7 +195,7 @@ export class IdentifierManagerFactory {
         }
         switch (algo) {
             case Algos.salty:
-                kargs = kargs as SaltyState;
+                kargs = kargs as SaltyKeyState;
                 return new SaltyIdentifierManager(
                     this.salter,
                     kargs.pidx,
@@ -214,7 +214,7 @@ export class IdentifierManagerFactory {
                     kargs.sxlt
                 );
             case Algos.randy:
-                kargs = kargs as RandyState;
+                kargs = kargs as RandyKeyState;
                 return new RandyIdentifierManager(
                     this.salter,
                     undefined,
@@ -229,7 +229,7 @@ export class IdentifierManagerFactory {
                     kargs.nxts
                 );
             case Algos.group:
-                kargs = kargs as GroupState;
+                kargs = kargs as GroupKeyState;
                 return new GroupIdentifierManager(
                     this,
                     kargs.mhab,
--- a/src/keri/core/keeping.ts
+++ b/src/keri/core/keeping.ts
@@ -9,11 +9,18 @@ import { Cipher } from './cipher';
 import { Diger } from './diger';
 import { Prefixer } from './prefixer';
 import { Signer } from './signer';
-import { HabState, State } from './state';
+import {
+    ExternState,
+    GroupState,
+    HabState,
+    RandyState,
+    SaltyState,
+    KeyState,
+} from './keyState';
 
 /** External module definition */
 export interface ExternalModuleType {
-    new (pidx: number, args: KeeperParams): Keeper;
+    new (pidx: number, args: IdentifierManagerParams): IdentifierManager;
 }
 
 export interface ExternalModule {
@@ -22,14 +29,14 @@ export interface ExternalModule {
     module: ExternalModuleType;
 }
 
-export type KeeperResult = [string[], string[]];
+export type IdentifierManagerResult = [string[], string[]];
 export type SignResult = string[];
 
-export interface KeeperParams {
+export interface IdentifierManagerParams {
     [key: string]: unknown;
 }
 
-export interface SaltyParams extends KeeperParams {
+export interface SaltyManagerParams extends IdentifierManagerParams {
     pidx: number;
     kidx: number;
     tier: Tier;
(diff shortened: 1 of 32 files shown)


kentbull:
Another goal of this PR is to get ready to add CESR 2.0 and KERI 2.0 support.

kentbull:
Good points Ed. I'll take a look at the signify-browser-extension code to see what exports it depends on.

kentbull:
@edeykholt item 3 of your notes above is addressed here: https://github.com/WebOfTrust/signify-browser-extension/pull/229

kentbull:
And @edeykholt , to your first point

> 1) reassurance that signify-browser-extension is not affected;

the results of a full text search on the signify-browser-extension repository for the names I changed here show that nothing in the browser extension depends on the names I changed here.

I will keep this in mind as I make further changes this week. If there are any refactorings that affect the browser extension then I will create a corresponding PR to that repo.

src/keri/core/keyState.ts
```diff
@@ -1,7 +1,7 @@
 import { Algos } from './manager';
 import { Tier } from './salter';
 
-export interface State {
+export interface KeyState {
```
lenkan, line 4: I feel this is a bit backwards. Isn't this more the IdentifierState?

Then the `SaltyState` below is more of a `SaltyKeyState`, i.e., the state for the specific salty key.

src/keri/core/manager.ts
```diff
@@ -450,22 +571,23 @@ export class Manager {
...
-        dcode = MtrDex.Blake3_256,
-        algo = undefined,
-        salt = undefined,
-        stem = undefined,
-        tier = undefined,
-        rooted = true,
-        transferable = true,
-        temp = false,
-    }: ManagerInceptArgs): [Array<Verfer>, Array<Diger>] {
+    incept(mgrIcpArgs: ManagerInceptArgs): [Array<Verfer>, Array<Diger>] {
```
lenkan, line 574: Imo, just `args` is sufficient here. It is obvious from the context that it is the arguments of the incept method of the manager. `mgrIcpArgs` is more confusing in my opinion.

src/keri/core/salter.ts
```diff
@@ -18,9 +21,24 @@ interface SalterArgs {
...
 
+    /**
+     * Creates a Salter from the provided raw salt bytes or generates a random salt if raw is not provided.
+     * Defaults to low security tier. Only supports Salt_128 salt type.
+     * @param raw
+     * @param code
+     * @param tier
+     * @param qb64
+     * @param qb64b
+     * @param qb2
```
lenkan, line 40: Are these docs actually correct? There is only one parameter to the constructor, and it is of type `SalterArgs`. Just to make sure this actually does what is intended.

lenkan: It might cause confusing intellisense in editors.

src/keri/core/manager.ts
lenkan: Same for the other methods here

src/keri/core/manager.ts
```diff
@@ -9,38 +9,107 @@ import { Cigar } from './cigar';
...
-    pubs: Array<string> = new Array<string>(); // list qb64 public keys.
+    /**
+     * List of fully qualified, Base64 encoded public keys.
+     */
+    pubs: Array<string> = new Array<string>();
 }
 
+/**
+ *
+ */
```
lenkan, line 112:
```suggestion
```

src/keri/core/keeping.ts
```diff
@@ -41,27 +48,33 @@ export interface SaltyParams extends KeeperParams {
...
+export interface GroupManagerParams extends IdentifierManagerParams {
     mhab: HabState;
 }
 
-export interface Keeper<T extends KeeperParams = KeeperParams> {
+/**
+ * Interface for KERI identifier (prefix) creation, rotation, and signing.
+ * @param T Type of the key keeper
+ */
+export interface IdentifierManager<
```
lenkan, line 65: Isn't this more of a `KeyManager` than an `IdentifierManager` ? You use it to create keys, which you intern use to create identifiers.

lenkan (commented):
Great work with all the cleanup and comments!

I added some minor nitpicking. But mostly, I am not sure about the `IdentifierManager` vs `KeyManager` naming. See the comment in the `keeping.ts` file. I read your rationale, but maybe I misunderstood.

src/keri/core/keeping.ts
kentbull: The reason I called it `IdentifierManager` is because that name communicates the broader scope of the class. An identifier can do incept, rotate, and sign. A key can sign, yet, not the other things. So that's why I picked IdentifierManager instead of KeyManager.

src/keri/core/manager.ts
kentbull: I added docs for PubPath. I started this blank comment before I knew what PubPath was used for. Now that I know then I added the docs.

src/keri/core/salter.ts
kentbull: good catch, this is fixed.

src/keri/core/keyState.ts
kentbull: The reason I called this `KeyState` is because that aligns with what KERIpy seems to use with things like [`KeyStateRecord`](https://github.com/WebOfTrust/keripy/blob/9e51b4ad80eb579d8223e29c7afcd6bb1c6c3535/src/keri/db/basing.py#L201) objects.

I refrained from changing SaltyState, RandyState, and GroupState since they seemed specific enough, yet on second thought it sounds good. I'll make that change.

kentbull:
@lenkan thanks for the comments! I made corresponding changes or added rationale for context.

2byrds (approved):
Thank you for responding to the feedback. I hope this achieves the alignment attended. Time will tell i suppose, but glad that we are trying :)

PR #311, opened 2025-02-12
fix: send exn for each recipient; no early return

### LINKED ISSUE #310: Bug: Exchanges.send() should not return early and instead execute for each recipient
### Version

0.3.0-rc1

### Environment

All

### Expected behavior

Exchanges.send() should send an exchange message to each recipient. 

### Actual behavior

Instead it returns early on the first recipient as shown in the code below.

```typescript
async send(
        name: string,
        topic: string,
        sender: HabState,
        route: string,
        payload: Dict<any>,
        embeds: Dict<any>,
        recipients: string[]
    ): Promise<any> {
        for (const recipient of recipients) {
            const [exn, sigs, atc] = await this.createExchangeMessage(
                sender,
                route,
                payload,
                embeds,
                recipient
            );
            return await this.sendFromEvents(
                name,
                topic,
                exn,
                sigs,
                atc,
                recipients
            );
        }
    }
```

### Steps to reproduce

The code shows the bug. The corrected implementation is shown below:
```typescript
async send(
    name: string,
    topic: string,
    sender: HabState,
    route: string,
    payload: Dict<any>,
    embeds: Dict<any>,
    recipients: string[]
): Promise<any[]> {  // Return an array of responses
    const results = [];  // Collect results for each recipient

    for (const recipient of recipients) {
        const [exn, sigs, atc] = await this.createExchangeMessage(
            sender,
            route,
            payload,
            embeds,
            recipient
        );
        
        const result = await this.sendFromEvents(
            name,
            topic,
            exn,
            sigs,
            atc,
            [recipient]  // Send individually
        );

        results.push(result);  // Store the response
    }

    return results;  // Return all results
}
```

kentbull:
Fixes #310 . The existing code would return early rather than execute for each recipient. This was okay from a functional perspective because of how KERIA acts as a delivery service for EXN messages. Yet, the way the code is written it indicates the for loop would execute for each member of the array, which is not how the code works because it returns early.

KERIA would take the list of recipients in the EXN message and then deliver all EXN messages as declared.

### DIFF
--- a/src/keri/app/exchanging.ts
+++ b/src/keri/app/exchanging.ts
@@ -79,6 +79,7 @@ export class Exchanges {
         embeds: Dict<any>,
         recipients: string[]
     ): Promise<any> {
+        const results = [];
         for (const recipient of recipients) {
             const [exn, sigs, atc] = await this.createExchangeMessage(
                 sender,
@@ -87,15 +88,17 @@ export class Exchanges {
                 embeds,
                 recipient
             );
-            return await this.sendFromEvents(
+            const result = await this.sendFromEvents(
                 name,
                 topic,
                 exn,
                 sigs,
                 atc,
-                recipients
+                [recipient]
             );
+            results.push(result);
         }
+        return results;
     }
 
     /**


iFergal:
I had attempted to fix this ages ago in #273 and never got the chance to revisit.

This doesn't work for end role auths as the notification won't appear unless they all sign the same exn, and putting individual member identifiers into each exn (`rp`) breaks this.

One approach might be to have a commitment on something internal to the exn and not the exn itself, but not sure how much that complicates other things.

The approach I discussed with @lenkan before was setting the `rp` field to the group identifier itself, and adjust KERIA calculate the other recipients to send it to. This also has the advantage of signing once at the edge, and one HTTP request.

iFergal:
Technically, we can just set the `rp` field to the group identifier right now, since I think it's the recipients parameter in the HTTP body that KERIA uses. But I believe the intention was to remove that parameter and always send to `rp`, but in this case KERIA would convert from `rp` -> `otherGroupMembers`.

PR #312, opened 2025-02-18
refactor: ked -> sad, ident -> proto

kentbull:
Aligns with KERIpy naming usage of the "sad" self-addressing data label and "proto" protocol label. The "ked" label was overused compared to KERIpy.

This PR slightly refactors the Serder class to set properties in the constructor and removes the .ked property for now.


kentbull:
Not sure what's up about the test coverage. No new code was added, just name changes.

iFergal:
I personally always use ked in keripy/keria

kentbull:
@lenkan 

> how did you come to the conclusion that ked was overused compared to keripy?

It was just visual inspection. `sad` is more generalized to the `Serder` base class in KERIpy, `ked` is specific to `SerderKERI`.

lenkan:
> It was just visual inspection. `sad` is more generalized to the `Serder` base class in KERIpy, `ked` is specific to `SerderKERI`.

Fair enough. So because the `Serder` in signify-ts is not specific to keri, it makes more sense to use the generic term `serializable attribute dictionary` rather than `key event dictionary` ?

lenkan (approved):

kentbull:
Yes, those were my thoughts. I do plan to reintroduce `.ked` with `SerderKERI` as I add a full CESR parser to SignifyTS. So if we end up wanting to use it then it will be here soon enough 😄

PR #313, opened 2025-03-04
fix: remove incorrect accept header for issue or revoke credential

iFergal:
The issue and revoke credential endpoints return an operation like most endpoints, so they should have an `application/json` content type.

The problem with the client specifying `application/json+cesr` (which `credentials().get(said)` _can_ do, but not by default) is the response body is lost if you raise e.g. a `falcon.HttpNotFound` exception. So if you tried to issue a credential with an unknown registry you would just see `HTTP POST /identifiers/issuer/credentials - 404 Not Found - `.


iFergal:
Not sure why the tests failed on the second commit. Seems like a flakey (very rarely) test, should be OK on a re-run. (I can't trigger)

2byrds:
> Not sure why the tests failed on the second commit, passing locally. Seems like a flakey (very rarely) test, should be OK on a re-run. (I can't trigger)

@iFergal I relaunched them but the validation action failed

PR #315, opened 2025-03-21
chore: remove some old and unmaintained dependencies

lenkan:
I wanted to upgrade typescript version and these are just some of the things that are blocking me.


lenkan:
#316 supersedes this if that one is accepted

PR #316, opened 2025-03-21
chore: upgrade typescript

lenkan:
This PR upgrades the typescript dependency to 5.8.

- The build process now simply transpiles the files and puts them in dist.
- Migrate from jest to vitest because I had a lot of issues upgrading with jest. Jest is pretty much abandoned project now anyway. Ideally, I would have wanted to move to node built in test runner, but then you need a way to load the typescript files, which is much easier if we move to node 22, or even 24 when that is available.
- Upgrade eslint to latest.

The reason for doing this is so that we can create a package that depends on signify-ts. The current build output from signify-ts does not play well with that. I can provide details later if anyone is interested. But in general, we should not bundle this library for publishing to npm.


2byrds (approved):
LGTM. Good to know about jest vs. vitest

PR #317, opened 2025-03-24
add npm publish script for easier npm publish

lenkan:
Tested here on my fork: https://github.com/lenkan/signify-ts/actions/runs/14033855540/job/39287547534

As an example, here is my package scope that I use for testing: https://www.npmjs.com/package/@lenkan/signify-ts?activeTab=versions

This could also be used to publish dev tags on push to main from the weboftrust repo. That way we can avoid the prepare script and other potential issues that comes with installing from github.


2byrds (approved):
LGTM

PR #318, opened 2025-03-24
simplify project folder structure

lenkan:
- Remove the workspace setting from package.json. See #228 for problems that can arise from that. Also, since we do not have the react app anymore, there is really no need for this repository to be a workspace.
- The test-integration files depend on the source files directly, so you will not need to build the src/ directory before you run the tests. Previously, if you forgot to build the project before running the tests, you would have an outdated `dist/` directory which could cause confusion.

The idea is to add more test suites. For example, testing in browser, smoke tests for the build output etc.

In the future, we could re-introduce the workspace if we want to publish multiple packages from this repository. For example `cesr-ts`.


lenkan:
> Nice, long overdue! Only thing on the naming for me is I feel like they are more e2e tests than integration.

I don't have a strong opinion for that. I thought of reserving e2e for tests running in a browser, but that could also be `test-browser`. Perhaps another PR for renaming to test-e2e if people think that is more appriopriate.

kentbull (approved):
Looks good to me. Approved ✅

PR #321, opened 2025-03-25
chore: remove buffer dependency

lenkan:
'Buffer' is a node.js-only API. So exposing it from signify-ts can cause confusion for people using it in browsers, even if we use a polyfill package. This PR removes all references to `Buffer`.

Also, note that `base64-js` was a dependency of the `buffer` module: https://github.com/feross/buffer/blob/5857e295f4d37e3ad02c3abcbf7e8e5ef51f3be6/package.json#L19, so now we just use the base64 encoding explicitly.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "Apache-2.0",
       "dependencies": {
         "@noble/hashes": "^1.3.2",
-        "buffer": "^6.0.3",
+        "base64-js": "^1.5.1",
         "ecdsa-secp256r1": "^1.3.3",
         "libsodium-wrappers-sumo": "^0.7.9",
         "mathjs": "^12.4.0",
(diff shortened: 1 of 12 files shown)


2byrds (approved):
lgtm

kentbull:
The Buffer library can be used in the browser via browserify. The Buffer API is user friendly enough that I recommend we switch to it by default. Specifically, this version: https://github.com/feross/buffer

kentbull:
From their readme, this package is intended for the browser:

lenkan:
Just as a side note. The Uint8Array will get toBase64 and fromBase64 functions going forward.

Typed arrays are natively supported by browsers. Buffer was introduced to node.js because there was no UInt8Array in the language at the time. I am not sure what operations you mean that would be simplified by using Buffer? But perhaps if you open upp a PR with the changes to inhale and exhale, it would be easier to grasp.

Another note is that if we decide to expose `Buffer` in the public API of signify-ts, you would force consumers of the API to import from the `buffer` package rather than just using the built in Uint8Array. For example


 ```typescript
 // Without buffer
 import { SignifyClient } from "signify-ts"
 const client = new SignifyClient()
 
 client.verifySomething(new Uint8Array([1, 2, 3]);
 
 // With buffer
 import { Buffer } from "buffer"
 import { SignifyClient } from "signify-ts"
 
 client.verifySomething(Buffer.from([1, 2, 3]));
 ```
 
 This can also lead to issues if the consumer and signify-ts has a dependency in different versions of the "buffer" package.

PR #336, opened 2025-07-02
refactor: remove redundant test helpers, and use them more

iFergal:
Separating this from https://github.com/WebOfTrust/signify-ts/pull/335 - I am planning on using `getOrCreateClient` to accept the auth mode (signed headers or ESSR) from the env, so we can add a second integration test run in the matrix.

### DIFF
--- a/test-integration/challenge.test.ts
+++ b/test-integration/challenge.test.ts
@@ -8,10 +8,7 @@ import {
 } from './utils/test-util.ts';
 
 test('challenge', async () => {
-    const [
-        client1,
-        client2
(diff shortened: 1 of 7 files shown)


docker-compose.yaml
```diff
@@ -10,7 +10,7 @@ x-python-env: &python-env
 
 services:
     vlei-server:
-        image: gleif/vlei
+        image: gleif/vlei:0.2.0
```
iFergal, line 13: see comment here: https://github.com/WebOfTrust/signify-ts/pull/335#discussion_r2180182330

PR #342, opened 2025-08-05
feat: Cache client resources

Sotatek-DukeVu:
Cache the instances of SignifyClient.


lenkan (commented):
Does this really have a performance hit?

If it has a measurable performance benefit, perhaps it can be simplified to:

```
private _identitiers = new Identifiers(this);
```

To avoid the nullchecks.

iFergal (approved):

Issue #341, opened 2025-07-31
Remove use of any in src/keri/app that isn't API response related


iFergal:
https://github.com/WebOfTrust/keria/issues/369 will provide types for a lot of the `src/keri/app` directory but some files do not depend on OpenAPI types from KERIA.

This should aim to remove the use of `any` from methods that are not API calls. This mostly applies to:
- `controller.ts`
- `habery.ts`
- Setup of agent in `controller.ts`

PR #344, opened 2025-08-14
chore: Remove use of any in src/keri/app that isn't API response related

Sotatek-DukeVu:
Issue: https://github.com/WebOfTrust/signify-ts/issues/341


kentbull:
This seems like a worthy PR. I'll review it if you re-open it.

PR #350, opened 2025-10-24
aiding.py type hints and auto-generated OpenAPI specs

Sotatek-Patrick-Vu:


src/keri/app/aiding.ts
```diff
@@ -73,6 +74,8 @@ export interface IdentifierInfo {
     name: string;
 }
 
+export type GroupMember = components['schemas']['GroupMember'];
```
iFergal: `GroupMembers`

test/core/manager.test.ts
```diff
@@ -750,7 +754,7 @@ describe('Manager', () => {
                 transferable: false,
                 windexes: [],
                 icp_dt: '2023-12-01T10:05:25.062609+00:00',
-            })
+            } as unknown as HabState)
```
iFergal: Can we update these tests to pass enough fields that we don't need to cast to `as unknown`?

src/types/keria-api-schema.ts
```diff
@@ -315,10 +332,246 @@ export interface components {
             pre: string;
             state: components['schemas']['CredentialState'];
         };
+        StateEERecord: {
+            /** @default 0 */
+            s: string;
+            /** @default  */
+            d: string;
+            br?: unknown[];
```
iFergal: Should these be optional? `ba` too?

Sotatek-Patrick-Vu: As discussed, we will change in KERIA to have `br: string[]`

kentbull (approved):

src/types/keria-api-schema.ts
iFergal: Will merge once we have the new PR to override that dataclass

iFergal (approved):
I'm going to merge this - we can fix the small `unknown` issue I mentioned in a subsequent PR.

PR #370, opened 2026-04-21
0.3.0

kentbull:
This is the v0.3.0 I released to NPM today. 

Upgraded the libsodium-wrappers-sumo to address a build problem. The `@types/libsodium-wrappers-sumo` is packaged with the core `libsodium-wrappers-sumo` as of v0.8.2 so we can drop the explicit dependency.

EDIT:
Also dropped Node 18 support (EoL Apr 30, 2025) - see below comment.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "signify-ts",
-  "version": "0.3.0-rc1",
+  "version": "0.3.0-rc2",
   "description": "Signing at the edge for KERI, ACDC, and KERIA",
   "keywords": [
     "keri",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "signify-ts",
-  "version": "0.3.0-rc2",
+  "version": "0.3.0",
   "description": "Signing at the edge for KERI, ACDC, and KERIA",
   "keywords": [
     "keri",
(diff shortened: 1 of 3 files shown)


kentbull:
I dropped Node 18 support as it required a custom test setup and vitest config. Node 18 is EoL as of April 30, 2025, and so we probably are okay dropping it.

If someone disagrees and wants us to keep Node 18 support then I'll add the following to our test setup. We just need to manually define the WebCrypto "crypto" global prop since Node 18 doesn't auto-load this, yet newer Node runtimes do.

For example:

in `test/setup-node-crypto.ts` we would put

```ts
import { webcrypto } from 'node:crypto';

if (!globalThis.crypto) {
    Object.defineProperty(globalThis, 'crypto', {
        value: webcrypto,
        configurable: true,
    });
}
```
and in `vitest.config.ts` we would put
```ts
import { defineConfig } from 'vitest/config';

export default defineConfig({
    test: {
        name: 'Unit tests',
        root: 'test',
        setupFiles: ['./setup-node-crypto.ts'],
        testTimeout: 10000,
    },
});

```

Simple enough if we want it. I figured dropping Node 18 support was a better idea.

package.json
```diff
@@ -64,7 +63,7 @@
     "@noble/curves": "^1.8.1",
     "@noble/hashes": "^1.3.2",
     "base64-js": "^1.5.1",
-    "libsodium-wrappers-sumo": "^0.7.9",
+    "libsodium-wrappers-sumo": "0.8.4",
```
lenkan: Usually you'd want consumers to be able to upgrade transitive dependencies. "^0.8.4" allows for that. Any particular reason this is locked here?

kentbull: I didn't know that, thanks for telling me. I'll update the PR.

lenkan (approved):

iFergal (approved):

PR #376, opened 2026-05-07
feat: add generic agent SSE signaling client

kentbull:
This adds the SignifyTS client counterpart to KERIA's SSE AgentSignals, broadcaster, SSE endpoint, and SSE iterable. This supports one or more Signify clients connecting to a given server Agent to get unique server sent events streams per SignifyClient controller instance.

This is in preparation for future event driven features including the did:webs integration, among others. The SSE event stream is also a convenient way for clients to switch from long-polling to observing events sent from an agent. Long-polling is still be supported as a fallback, though this allows SSE to be the default future communication pathway from the agent to a client. This will significantly help with both bandwidth and performance on highly used Agency servers.

This PR should be merged along side WebOfTrust/keria #435 https://github.com/WebOfTrust/keria/pull/435

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,19 @@ Typescript source files needs to be transpiled before running scripts or integra
     console.log(actualSignifyClient);
     ```
 
+### Agent signaling
+
+KERIA can publish transient server-sent events for the connected agent at
+`GET /signals/stream`. SignifyTS exposes this generic channel through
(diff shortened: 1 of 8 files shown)


kentbull:
https://github.com/WebOfTrust/signifypy/pull/145 is the SignifyPy correlate to this.

Issue #378, opened 2026-05-12
ondex calculated incorrectly from new rotation instead of prior establishment event


kentbull:
### Version

0.3.0

### Environment

All

### Expected behavior

Doing rotations for multisig groups where simultaneously an existing member is dropped and a new member added then the rotation should succeed.

### Actual behavior

When doing a multisig rotation where a member was both removed and a new one added then the rotation fails. This is because the `ondex`, the index of the precommitted key used in threshold satisfaction checks, was incorrectly being derived from the proposed new rotation event rather than the prior establishment event's next key digest list in the `n` attribute.

### Steps to reproduce

1. Make a multisig group with members A, B, and C.
2. Make a rotation dropping C and adding D in the same rotation.
3. The bug manifests as a failed rotation.

PR #379, opened 2026-05-12
critical fix: #378 calculate ondex from prior est evt, not proposed one

kentbull:
## Problem

Failed multisig rotation where a new member is added and a new one is dropped during the same rotation. The root cause is the incorrect calculation of the `ondex`, the index of the prior next key digest, from the proposed rotation event, from `rstates` passed in to `GroupIdentifierManager` in the constructor. `rstates` is for the proposed event, not the prior establishment event.

## Solution

Pull the `n` next keys (prerotated committed keys) from the prior establishment event and calculate the `ondex` values from that rather than the `rstates` from the proposed event.

## Background

KERIpy validates ondexes, which are the pre-committed next key digest indexes, from the prior establishment event, not the new, proposed establishment event.

The reason we haven't hit this in the past is because all of the existing tests never swapped in a new member in the index/ondex of a prior member. We always had a clean addition of a new member and then a separate operation for the removal of the to-be-deleted member.

Essentially, we didn’t hit it because the existing tests never exercise departing-current-signer signs while absent from proposed rstates (the next set of pre-committments).

See #378 for an example rotation that reproduces this bug.


kentbull:
This also adds a fair bit of documentation to code that was poorly documented.

src/keri/core/keeping.ts
```diff
@@ -216,7 +217,8 @@ export class IdentifierManagerFactory {
                 undefined,
                 undefined,
                 aid.group.keys,
-                aid.group.ndigs
+                aid.group.ndigs,
+                aid.state?.n ?? aid.group.ndigs
```
iFergal: From so far as I can tell, `gdigs` is initialised as the current next key digests. They are only updated later when you call `.rotate`.

Can we not just directly use `this.gpndigs = this.gdigs;` in the constructor?

src/keri/core/keeping.ts
```diff
@@ -742,13 +816,87 @@ export class GroupIdentifierManager implements IdentifierManager {
...
+
+        return -1;
+    }
+
+    /**
+     * Extracts the serialized event ilk for local signing decisions. Parsing is
+     * intentionally conservative: unknown or unparsable payloads fall through to
+     * the non-rotation signing path.
+     */
+    private eventIlk(ser: Uint8Array): string | undefined {
```
iFergal: Would it be cleaner to have a `rotated` argument that determines this for `sign`, like keripy's `GroupHab`?

src/keri/core/keeping.ts
```diff
@@ -742,13 +816,87 @@ export class GroupIdentifierManager implements IdentifierManager {
...
+                // precommitted key digest, because prior `n` authorizes this
+                // event's current signer set.
+                pni = this.priorNextIndexForKey(key);
+                ondexSource = 'prior next digest list';
+                break;
+            default:
+                // Inception and non-rotation group signatures do not satisfy a
+                // prior-next rotation threshold. Preserve the existing same-list
+                // behavior for their `ondex` value.
+                pni = this.gdigs!.indexOf(ndig);
```
iFergal: Shouldn't it actually be undefined in this case? (Not sure if that breaks anything, but would seem more correct)

test-integration/multisig-prior-next-ondex.test.ts
```diff
@@ -0,0 +1,205 @@
...
+} from './utils/multisig-utils.ts';
+
+const NOTIFICATION_WAIT = {
+    maxRetries: 30,
+    minSleep: 500,
+    maxSleep: 500,
+    timeout: 15_000,
+};
+
+test('3-of-3 multisig replacement rotation signs with prior next ondexes', async () => {
```
iFergal: Is it worth folding this into `multisig-join`?

Also, perhaps we just frame this test as "can replace a group member" rather than mentioning ondex specifics at integration test level

src/keri/core/keeping.ts
kentbull: > Can we not just directly use this.gpndigs = this.gdigs; in the constructor?

From within `GroupIdentifierManager` then `this.gdigs` gets set from the `rstates` values, which are not the **current next key digests** but instead are the **proposed next rotation key digests**.

They are set to the wrong value when GroupIdentifierManager is constructed when it calls `this.client.manager!.get(hab):` (shown [here](https://github.com/WebOfTrust/signify-ts/blob/main/src/keri/app/aiding.ts#L375))

So, no, we cannot directly use `this.gdigs`. We need both gdigs to represent the proposed rotation event's next keys and we also need to keep track of the prior establishment event's next key digests.

src/keri/core/keeping.ts
kentbull: There is a bug in my code, that I am fixing, though the `default` case here currently handles inception, delegated inception, and interaction, which should have a pni set to None to align with KERIpy's GroupHab.sign. I am making that adjustment now.

src/keri/core/keeping.ts
iFergal: Hmm, but `this.client.manager!.get(hab);` will fetch the current state of the group from KERIA, so it hasn't been rotated at this point. The call [further down](https://github.com/WebOfTrust/signify-ts/blob/main/src/keri/app/aiding.ts#L381) updates it from the current next key digests to the proposed next rotation key digests, _after_ the `GroupIdentifierManager` has been initialised.

kentbull: You are right, I was wrong to say that the `this.client.manager!.get(hab)` populates the `states` and `rstates` because they are both `undefined` after constructor initialization. They are later set on the `.rotate()` call, though my comment still stands that we cannot directly use `this.gdigs` in the constructor because the `rstates` value is empty.

Are you getting at the possibility that we could pass in the `rstates` from the Hab state we get from KERIA? That could work if we were to not overwrite that state in `.rotate()` below, though we'd still want to track both sets of pre-committed keys, the prior next one as well as the proposed next set, so we would end up with two variables in GroupIdentifierManager anyway, one for the prior next set and one for the proposed next set.

src/keri/core/keeping.ts
kentbull: It might be. I'll take a look at that. It pays to mimic KERIpy where we can.

src/keri/core/keeping.ts
```diff
@@ -443,14 +445,12 @@ export class SaltyIdentifierManager implements IdentifierManager {
...
-                let o = 0;
+                let o: number | undefined = 0;
                 if (ondices != undefined) {
                     o = ondices![j];
                     if (
-                        (o == undefined ||
-                            (typeof o == 'number' &&
-                                typeof o != 'number' &&
-                                o >= 0))!
+                        o !== undefined &&
```
kentbull: This is just some cleanup of conditionals I noticed needed to happen, going from double equals to triple equals for clarity. Similar changes are made below.

src/keri/core/keeping.ts
iFergal: `this.client.manager!.get(hab)` uses `hab.group.keys` and `hab.group.ndigs` instead of `states` and `rstates`.

So they map directly to `gkeys` and `gdigs`:
```
        this.gkeys = states?.map((state) => state['k'][0]) ?? keys;
        this.gdigs = rstates?.map((state) => state['n'][0]) ?? ndigs;
```

So in this case, `gdigs` will contain the current next key digests when the instance of the class is created. When `rotate` is called, `gdigs` will be updated to the proposed next rotation key digests.

So, my proposal is `this.gpndigs = this.gdigs;` in the constructor will set `gpndigs` to the current next key digests. This works because `rotate` doesn't update `gpndigs`.

test-integration/multisig-prior-next-ondex.test.ts
kentbull: I like that idea, yet since the dual indexing, index + ondex, is a challenging enough of a concept for newcomers it seems valuable to have a dedicated file with hints in the labels that help people learn and understand the signature and threshold validation scheme.

src/keri/core/keeping.ts
kentbull: I like your proposal. I simplified the constructor to reflect this and all the tests pass, so we're good.

src/keri/core/keeping.ts
kentbull: I just refactored to add the `rotated=True` and it was cleaner.

src/keri/core/keeping.ts
kentbull: @iFergal you were correct and I refactored the code to have `pni` be `undefined` for all non rotation events.

src/keri/core/keeping.ts
```diff
@@ -649,14 +658,12 @@ export class RandyIdentifierManager implements IdentifierManager {
...
```
kentbull: conditionals cleanup like mentioned above.

src/keri/core/keeping.ts
```diff
@@ -734,21 +813,79 @@ export class GroupIdentifierManager implements IdentifierManager {
...
-        const pni = this.gdigs!.indexOf(ndig); // pni = prior next index (from last establishment event)
+        if (csi < 0) {
+            throw new Error(
+                `Group signing key ${key} is not present in current group signing keys.`
+            );
+        }
+
+        let pni: number | undefined; // prior next index exposed as signature `ondex`
+
+        if (rotated) {
```
kentbull, line 839: This `if (rotated)` is just like how KERIpy does GroupHab.sign.

src/keri/core/manager.ts
```diff
@@ -1086,19 +1086,22 @@ export class Manager {
                     i = idx;
                 }
 
-                let o;
+                let o: number | undefined;
                 if (ondices != undefined) {
                     o = ondices[idx];
-                    if (o <= 0) {
+                    if (
```
kentbull: conditionals cleanup like mentioned above.

test-integration/modules/bip39_shim.ts
```diff
@@ -126,14 +127,12 @@ export class BIP39Shim implements IdentifierManager {
                 } else {
                     i = j;
                 }
-                let o = 0;
+                let o: number | undefined = 0;
                 if (ondices != undefined) {
                     o = ondices![j];
                     if (
-                        (o == undefined ||
```
kentbull, line 133: conditionals cleanup like mentioned above.

test-integration/multisig-prior-next-ondex.test.ts
iFergal: Personally I don't think newcomers should be exposed to those details - they should just know how to replace users in a group. Anyway, in the long term newcomers shouldn't need to reference integration tests to learn how to use Signify so happy to keep.

src/keri/core/keeping.ts
```diff
@@ -416,7 +425,8 @@ export class SaltyIdentifierManager implements IdentifierManager {
         ser: Uint8Array,
         indexed = true,
         indices: number[] | undefined = undefined,
-        ondices: number[] | undefined = undefined
+        ondices: Array<number | undefined> | undefined = undefined,
```
iFergal, line 432: I'm not sure why we changed from `number[]` to `Array<number | undefined>`. In what scenario is something like `[undefined, undefined, undefined]` being passed instead of `undefined`?

src/keri/core/keeping.ts
iFergal: Does this change come from setting `pni` to undefined?

Also, we can remove `typeof o !== 'number'` I think to simplify. `Number.isInteger(undefined)` etc will return false and not error

test-integration/multisig-prior-next-ondex.test.ts
kentbull: fair

src/keri/core/keeping.ts
kentbull: No, this change came from me noticing the ugliness of the type checking and wanting to make it better.

Yes, I agree with your simplifications. I'll add them.

src/keri/core/keeping.ts
kentbull: The reason it is `<number | undefined>` is because I closely mimicked KERIpy's `[None]` array passing of `pni` to mhab.sign, like below:

```python
return (self.mhab.sign(ser=ser,
                               verfers=[merfer],
                               indexed=indexed,
                               indices=[csi],
                               ondices=[pni]))
```                               

When `pni` is `None` in this instance then ondices is set to `[None]` in Python, and the closest analog in our case is `[unknown]`.

test-integration/multisig-prior-next-ondex.test.ts
kentbull: the test is now moved to multisig-join

src/keri/core/manager.ts
kentbull: `Number.isInteger()` typecheck now added`Number.isInteger()` typecheck now added

test-integration/modules/bip39_shim.ts
kentbull: `Number.isInteger()` typecheck now added

src/keri/core/keeping.ts
iFergal: OK, essentially forced by my request for `pni = undefined;`, got it.

Happy to merge to not inflict changes elsewhere but it'd be more natural in TS to have `pni ? [pni] : undefined` or something because TS gives you checks for undefined at the transpiler level. `Array<number | undefined> | undefined = undefined,` isn't the best fit in this language.

iFergal (approved):

src/keri/core/keeping.ts
kentbull: It looked ugly to me, too. I'll clean this up. Good point.

daidoji (approved):

PR #381, opened 2026-05-26
Types for e, s, r accomodate SAIDs and blocks.

daidoji:
Changed type specifications to allow for the fact that these fields can accomodate SAIDs and blocks.

### DIFF
--- a/src/keri/app/credentialing.ts
+++ b/src/keri/app/credentialing.ts
@@ -356,11 +356,22 @@ export class Credentials {
 
         const keeper = this.client.manager.get(hab);
 
-        const [, subject] = Saider.saidify({
-            d: '',
-            ...args.a,
-            dt: args.a.dt ?? new Date().toISOString().replace('Z', '000+00:00'),
-        });
+        let subject;
+        let dt: string;
+        if (typeof args.a === 'string') {
+            subject = args.a;
+            dt = new Date().toISOString().replace('Z', '000+00:00');
+        } else if (typeof args.a !== 'string') {
+            dt =
+                args.a.dt ?? new Date().toISOString().replace('Z', '000+00:00');
+            [, subject] = Saider.saidify({
+                d: '',
+                ...args.a,
+                dt: dt,
+            });
+        } else {
+            throw new Error('Data section must be valid CredentialData');
+        }
 
         const [, acdc] = Saider.saidify({
             v: versify(Protocols.ACDC, undefined, Serials.JSON, 0),
@@ -381,7 +392,7 @@ export class Credentials {
             i: acdc.d,
             s: '0',
             ri: args.ri,
-            dt: subject.dt,
+            dt: dt,
         });
 
         const sn = parseInt(hab.state.s, 16);
--- a/src/keri/app/credentialing.ts
+++ b/src/keri/app/credentialing.ts
@@ -80,23 +80,23 @@ export interface CredentialData {
     /**
      * Registry id.
      */
-    ri?: string;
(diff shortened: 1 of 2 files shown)


src/keri/app/credentialing.ts
```diff
@@ -80,23 +80,23 @@ export interface CredentialData {
     /**
      * Registry id.
      */
-    ri?: string;
+    ri?: string; // Note: this is rd in the approved ACDC spec.
     /**
      * Schema id
      */
-    s?: string;
+    s?: string | { [key: string]: unknown }; // Either the SAID of the schema or the schema block itself.
```
iFergal, line 87: This type is used as an argument to issue a new credential, not as the type that is returned from list/get (`export type CredentialResult = components['schemas']['Credential'];`)

I presume when issuing we must supply everything fully expanded

daidoji: Yeah, I mean that might be an unintentional regression, we originally came across this when updating signify-ts on old code that accommodated passing the SAID of a rule directly like `r: SAID _OF_RULE` 

https://github.com/WebOfTrust/signify-ts/blob/0.2.0/src/keri/app/credentialing.ts#L68

It doesn't seem like it should be a rule that we MUST pass the uncompacted version of a credential when issuing as both forms are legal in the ACDC spec.  A user might want to hide uncompacted forms of credentials from prying eyes until necessary for graduated disclosure.

It does make the typing more honky though for sure.

iFergal: > It doesn't seem like it should be a rule that we MUST pass the uncompacted version of a credential when issuing as both forms are legal in the ACDC spec. A user might want to hide uncompacted forms of credentials from prying eyes until necessary for graduated disclosure.

Hmm, but if you don't supply it at issuance time, how do we (practically) uncompact it later for graduated disclosure?

I would've thought we always issue uncompacted (since this is a local thing aside from registry anchoring), and then compact when disclosing via IPEX.

--

Though, rules well may be well known so I see the advantage there

daidoji: I think that's the crux of the matter in ACDC.  The uncompacted and compacted credential are for all means and purposes the same credential when it comes to ACDC spec (at least in terms of their identity the credential top level SAID).  Its out of scope to the spec (and these tools) which version of the credential is issued/granted/presented whatever.  Its up to the verifier as a business decision whether or not to accept that credential.  

So the manner of graduated disclosure also exists at a higher level.  If I want to do that as the issuer I have to provide for that mechanism outside the IPEX process itself.

iFergal: Right but this PR adjusts `a` to allow it to be compacted. If I issue compacted as a Signify client, and later want to engage in a graduated disclosure and eventually reveal the uncompacted `a`, I need to externally fetch it rather from KERIA as a DB.

Anyway, if it works e2e then I guess no issue updating the type, but perhaps we can extend the integration tests to cover that?

daidoji: Yeah, that's a more difficult matter.  I've never done graduated disclosure myself with KERIA or even keripy and don't know exactly what the implementations support.  One imagines some future tool that would allow one to call something like `disclose_only(credential, mask)` in the future to take any given credential and display some compactified version of it and then maybe some tracking in an implementation to see which compactified versions they've actually been presented/verified/issued by another agent.  

I don't even know if keripy supports this today, much less keria.  

I will add some tests though for the alteration I had to make to the implementation though to accommodate the type change.

PR #382, opened 2026-05-26
refactor: clean up index typing with Ondex tiny type

kentbull:
Adds the `Ondex` type to clean up `Array<number | undefined>` to allow `Ondex[]` where some entries in an ondex array could be values and some can be undefined.

### DIFF
--- a/src/keri/core/keeping.ts
+++ b/src/keri/core/keeping.ts
@@ -91,7 +91,7 @@ export interface IdentifierManager<
         ser: Uint8Array,
         indexed?: boolean,
         indices?: number[],
-        ondices?: Array<number | undefined>,
+        ondices?: Ondex[],
         rotated?: boolean
     ): Promise<SignResult>;
 }
@@ -429,7 +429,7 @@ export class SaltyIdentifierManager implements IdentifierManager {
         ser: Uint8Array,
         indexed = true,
         indices: number[] | undefined = undefined,
-        ondices: Array<number | undefined> | undefined = undefined,
+        ondices: Ondex[] | undefined = undefined,
         _rotated?: boolean
     ): Promise<SignResult> {
         const signers = this.creator.create(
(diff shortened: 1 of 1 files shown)


src/keri/core/keeping.ts
```diff
@@ -818,7 +823,7 @@ export class GroupIdentifierManager implements IdentifierManager {
         ser: Uint8Array,
         indexed: boolean = true,
         _indices: number[] | undefined = undefined,
-        _ondices: Array<number | undefined> | undefined = undefined,
+        _ondices: Ondex[],
```
iFergal: But in the last PR merged, we kept `[pni]` so this isn't correct? https://github.com/WebOfTrust/signify-ts/pull/379/changes#r3235945270

kentbull: What we decided in the last PR was that the array as a whole cannot be undefined, so the old default of `= undefined` is inaccurate because ondexes should always be either:
1. an empty array, an array of `undefined` values per index (when ondexes are ignored)
2. an array with a mix of ondex values and `undefined` (when some ondexes are ignored)
3. an array of ondex values (when all ondexes are used and none a re ignored)

So, my thinking is that an Ondex array type is accurate here because an individual Ondex is `number | undefined`, meaning each value in an `Ondex[]` array type would be either `number` or `undefined`. Thus, `[pni]` is kept and each `pni` value can be `number` or `undefined`.

Is that how you saw it or did you see it differently?

iFergal: Ah, sorry I misread this PR - you're right.

What about for Salty, and the base `IdentifierManager` definition?

kentbull: Good catch, I was wondering the same thing. I just added a new commit with those two updated.

iFergal (approved):

PR #383, opened 2026-05-27
feat: NPM Trusted Publishing

kentbull:
Uses the npm-publish GitHub Environment along with GitHub Actions to perform trusted publishing to the npmjs.com NPM package registry using OIDC with scopes constraining the authorization to the local GitHub Actions runner for a given CI run.

### DIFF
--- a/.github/workflows/publish-npm.yml
+++ b/.github/workflows/publish-npm.yml
@@ -9,26 +9,28 @@ on:
           - dev
           - latest
         default: dev
-        description: Npm dist tag
+        description: NPM dist tag
 jobs:
   publish:
     name: Publish NPM
     permissions:
-        contents: write
+      contents: read
+      id-token: write
+    environment: npm-publish
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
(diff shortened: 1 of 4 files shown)


daidoji (approved):
Nice.

.github/workflows/publish-npm.yml
```diff
@@ -9,26 +9,28 @@ on:
...
     permissions:
-        contents: write
+      contents: read
+      id-token: write
+    environment: npm-publish
     runs-on: ubuntu-latest
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: actions/checkout@v6
```
iFergal, line 23: Not really for this PR, but I think we should start using explicit githashes for actions to be a bit more safe from supply chain attacks

iFergal (approved):
Overall LGTM, we should test with a dev release before merging though

.github/workflows/publish-npm.yml
kentbull: I agree on git hashes. I'll change to that in the next PR.

Issue #384, opened 2026-05-28
keria-api-schema enums and types not exported by default.


daidoji:
### Feature request description/rationale

Not sure if anyone else wants this but currently there are many types generated by the automated keria generated types from schema that aren't exported in the main signify-ts package.  It would be nice if these were exposed for runtime checking.

https://github.com/WebOfTrust/signify-ts/blob/main/src/types/keria-api-schema.ts

kentbull:
You can use them by importing the `components` interface and then using string property accessors like so:
```ts
components['schemas']['ACDC_V_1']
```

See [here](https://github.com/WebOfTrust/signify-ts/blob/f9b223d50e8f7476260ee9502afe48a67f795652/src/keri/app/exchanging.ts#L30)

daidoji:
@kentbull this works from within signify-ts but what about using signify-ts as an external dependency.   Is there an incantation to do it via the signify-ts package?

daidoji:
Closing as fixed.

PR #385, opened 2026-05-28
Adds generated api types to default exports

daidoji:
For discussion, these types and enums are hard to get at but are helpful when developing and using this library as a dependency.  This change allows these types to be exposed in the default import via the KeriaApi namespace.

A solution to #384

### DIFF
--- a/src/exports.ts
+++ b/src/exports.ts
@@ -42,3 +42,5 @@ export * from './keri/core/verfer.ts';
 export * from './keri/core/keyState.ts';
 
 export * from './keri/end/ending.ts';
+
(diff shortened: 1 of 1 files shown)


iFergal (approved):
Ideally they are re-exported where it makes sense (they are in places) but I'm happy to export them all like this for now!

daidoji:
@iFergal Awesome, yeah this PR was mainly for discussion.  Yeah you guys have done a lot of great work with the typing, and maybe there's a bigger meta-discussion about the typing strategy that I'm outside of but it def seems like we could still make things a lot easier for the end user.  Glad to contribute to the effort where I can.

kentbull (approved):

PR #388, opened 2026-06-04
Cover late-join multisig registry recovery

kentbull:
References WebOfTrust/keria#316.

This adds an integration test for the late-join multisig registry gap from KERIA #316. The test creates an A/B multisig registry, issues a credential to make registry/TEL CESR export available through credential export, rotates C into the group, confirms C does not initially list the registry, imports the credential CESR, renames the registry, and verifies C can read/export the credential.

It also verifies the recovered registry remains usable by issuing a new credential with A, B, and C after the join. The shared multisig credential helper now returns the issued credential SAID so tests can track the exact request without local wrappers.

Validation:
- npm run build
- Targeted SignifyTS integration: multisig-late-join-registry-repro.test.ts, 1 passed
- git diff --check

### DIFF
--- a/test-integration/multisig-late-join-registry-repro.test.ts
+++ b/test-integration/multisig-late-join-registry-repro.test.ts
@@ -1,0 +1,1104 @@
+import assert from 'assert';
+import { test } from 'vitest';
+import signify, {
+    CreateIdentiferArgs,
+    CredentialData,
+    CredentialResult,
+    EventResult,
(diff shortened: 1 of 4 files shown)


kentbull:
Related PRs for the same WebOfTrust/keria#316 late-join multisig registry work:

- WebOfTrust/keria#439
- WebOfTrust/signifypy#149

Issue #392, opened 2026-06-09
Integration tests fail against dockerhub's KERIA 0.4.0/latest image


daidoji:
### Version

0.4.0 (or latest main)

### Environment

linux, 25.2.1

### Expected behavior

Integration tests should pass against the latest KERIA version.

### Actual behavior

Integration tests fail when run against `KERIA_IMAGE=latest` or `KERIA_IMAGE=0.4.0`

```
$ npm run test:integration
...
 ❯  Integration tests  multisig-vlei-issuance.test.ts (1 test | 1 failed) 14616ms
   × multisig-vlei-issuance 14612ms

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL   Integration tests  multisig-vlei-issuance.test.ts > multisig-vlei-issuance
Error: HTTP POST /oobis - 400 Bad Request - {"title": "400 Bad Request", "description": "invalid OOBI request body, either 'rpy' or 'url' is required"}
 ❯ SignifyClient.fetch ../src/keri/app/clienting.ts:238:19
    236|             const error = await res.text();
    237|             const message = `HTTP ${method} ${path} - ${res.stat…
    238|             throw new Error(message);
       |                   ^
    239|         }
    240|         const isSameAgent =
 ❯ Oobis.resolve ../src/keri/app/coring.ts:97:21
 ❯ getOrCreateContact utils/test-util.ts:279:31
 ❯ run multisig-vlei-issuance.test.ts:156:5

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯


 Test Files  1 failed (22)
[...]

PR #397, opened 2026-06-24
tests: Fix the integration tests to work with 0.4.0

kentbull:
- Fixes the docker compose file to use "keria" rather than "agent" as the agency name.
- Upgrades to KERIA 0.4.0 for the docker compose file, and KERIpy 1.2.13 for witnesses

Cleans up multisig notification expectations during notification wait loops by making them strict according to the following rules:
- `/multisig/exn` or other `/multisig/*` messages should only ever create notifications for non-initiators. 
  - Prior to #401 then `hby.exc`, the default exchanger, only had generic Exchanger behavior for EXNs submitted through `/identifiers/{name}/exchanges` which did not deduplicate coordination notifications.
  - The important changes are:
    - `/multisig/exn` notifications are now deduplicated since Multiplexor records it as approved/submitted and dedupes it.
    - `/multisig/exn` messages are properly indexed so they are searchable by the edge client.
    - This changed from 0.2.0-rc2 to 0.4.0 because #401, correctly routed Exchange messages to the Agent's exchanger (`agent.exc`), and eventually to the `Multiplexor.add` function, which deduplicates exn notifications by source/sender being local or not.
- `assertNoNotifications` added to make explicit when notifications should not be present. This shapes our integration tests around the specific notification behavior of KERIpy and KERIA so we know when something changes.
- `NotificationOptions` extends RetryOptions with `said` to support specific retrieval of exn messages by send for consistent notification selection during IPEX Grant operations and `minCount` to support batch handling of `/multisig/rpy` notifications for endpoint role notification handling.
- `matchesNotification` to clean up notification filter reuse after adding `said` to the filter.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -33,16 +33,16 @@ jobs:
           npm cache clean --force
           npm set registry https://registry.npmjs.org/
           npm ci
-      
+
       - name: Check formatting
-        run: npm run pretty:check 
-      
+        run: npm run pretty:check
+
       - name: Lint
         run: npm run lint
 
       - name: Build
         run: npm run build
- 
+
       - name: Test
         run: npm test -- --coverage=true
 
@@ -57,7 +57,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        keria-version: ['0.2.0-rc2', 'latest']
+        keria-version: ['0.4.0', 'latest']
         node-version: ['20']
     env:
       KERIA_IMAGE_TAG: ${{ matrix.keria-version }}
--- a/test-integration/credentials.test.ts
+++ b/test-integration/credentials.test.ts
@@ -9,12 +9,14 @@ import {
 import { resolveEnvironment } from './utils/resolve-env.ts';
 import {
     assertNotifications,
+    assertNoNotifications,
     assertOperations,
     createAid,
     getOrCreateClients,
     getOrCreateContact,
     markAndRemoveNotification,
     resolveOobi,
+    waitAndMarkNotification,
     waitForNotifications,
     waitOperation,
 } from './utils/test-util.ts';
@@ -52,6 +54,12 @@ let legalEntityAid: Aid;
 let applySaid: string;
 let offerSaid: string;
 let agreeSaid: string;
+let qviGrantSaid: string;
+let presentationGrantSaid: string;
+let leGrantSaid: string;
+let qviAdmitSaid: string;
+let presentationAdmitSaid: string;
+let leAdmitSaid: string;
 
 beforeAll(async () => {
     [issuerClient, holderClient, verifierClient, legalEntityClient] =
@@ -257,6 +265,7 @@ test('single signature credentials', { timeout: 90000 }, async () => {
             recipient: holderAid.prefix,
             datetime: dt,
         });
+        qviGrantSaid = grant.said;
 
         const op = await issuerClient
             .ipex()
@@ -264,6 +273,9 @@ test('single signature credentials', { timeout: 90000 }, async () => {
                 holderAid.prefix,
             ]);
         await waitOperation(issuerClient, op);
+        await waitAndMarkNotification(issuerClient, '/exn/ipex/grant', {
+            said: qviGrantSaid,
+        });
     });
 
     await step(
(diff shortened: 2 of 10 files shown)


kentbull:
Fixes #392

.github/workflows/main.yml
```diff
@@ -57,7 +57,7 @@ jobs:
     strategy:
       matrix:
         os: ['ubuntu-latest']
-        keria-version: ['0.2.0-rc2', 'latest']
```
daidoji, line 60: As long as we have the test matrix shouldn't we test against all supported versions as long as those versions are deployed? 

Or like what is the backwards compatibility guarantee we give on supported versions?

kentbull: That’s not a bad idea, though notification behaviors changed between versions and so running version specific tests would be complicated. I would either have to loosen the notification assertions or write version specific tests.

Right now there are no supported versions as GLEIF does not offer support contracts. We do our best to provide limited support for the last few versions, yet our allocated resources are currently few for version specific support.

It’s mostly discipline and whatever issue gets to the top of the pile that constitutes support at the moment.

daidoji: Well what I mean is that we know currently that `0.2.0-rc2` passes the tests, maybe we just add `0.4.0`.  It doesn't add much of a burden at the moment.  If our contract as maintainers aren't to a specific version then why not just always use `latest`?

kentbull: 0.2.0-rc2 does not pass the tests with the strict, new notification assertions. I’ll take a look and see the delta of what we would need to change to allow supporting 0.2.0-rc2 in the tests. It would be good to keep it in the test matrix for as long as practical.

kentbull: @daidoji after looking into it maintaining test assertion compatibility with 0.2.0-rc2 is not worth it as we'd have to change test fixtures to switch based on version tag.

Here's the difference:
- v0.2.0-rc2 and prior surface redundant multisig notifications to multisig operation initiators when receiving Exchange messages with routes like `/multisig/icp` or `/multisig/ixn` because they failed to route multisig notifications through Multiplexor, which has de-duplication of multisig notifications for initiators built in at the KERIpy level.
- 0.4.0 includes Fergal's https://github.com/WebOfTrust/keria/pull/401 fix to set `exc=agent.exc` for the Agent's Habery parser, which removes duplicate notifications for multisig initiators.

I recommend upgrading to 0.4.0.

kentbull: There is a changelog to see what changed between the versions: https://github.com/WebOfTrust/keria/blob/main/changelog.md

daidoji: Sure if its too much work we don't have to.  I think eyeballing the changelog is a bit more difficult to detect regressions than test suites.  That being said we're migrating to 0.4.0 so I guess others will just have to be cognizant.

daidoji (approved):

.github/workflows/main.yml
kentbull: Yeah, eyeballing it from a changelog is definitely more difficult to detect, and we don't even have good changelog discipline yet. We recently added the changelog, and it is not yet as itemized as I would want. Each line item should, generally speaking, be tied to one or more PRs.

We've got a ways to go.

iFergal: My change in `exchanging.py` introduces _more_ notifications. For example, if I control a single-sig AID and submit an exn, it will be parsed with a properly loaded `exc`, which means for known routes it will typically create a notification.

This might be seen as annoying but the benefit is we have proper route-based validation for exns we are submitting. It caused a problem with the tests though since they often assert that we have no unread notifications at the end of the test.

test-integration/credentials.test.ts
```diff
@@ -257,13 +265,17 @@ test('single signature credentials', { timeout: 90000 }, async () => {
...
 
         const op = await issuerClient
             .ipex()
             .submitGrant(issuerAid.name, grant, gsigs, gend, [
                 holderAid.prefix,
             ]);
         await waitOperation(issuerClient, op);
+        await waitAndMarkNotification(issuerClient, '/exn/ipex/grant', {
+            said: qviGrantSaid,
+        });
```
iFergal, line 278: An alternative might be to just ignore any related to exns we send from AIDs we control in the final count, rather than marking them at each step due to https://github.com/WebOfTrust/keria/pull/401 - this would keep the tests a bit easier to follow imo. On the other hand, it's less obvious to newcomers that these extra notifications are created. Thoughts?

kentbull: I wondered the same, maybe just ignore the notifications. However, I've seen QVIs learn the KERIA and SignifyTS API expectations by reading signify-ts integration tests. And, notifications are something that is less obvious to newcomers and because of that I favor explicit notification assertions in order to teach newcomers about how notifications work.

Many QVIs at GLEIF start building their systems, whether AI-driven or hand-coded, from the signify-ts integration tests. So strict notification behaviors in the integration tests act as a form of active documentation of our API design and expectations regarding notifications.

While ignoring related exns sent to AIDs we control could make the tests shorter and easier to follow due to less verbosity, they would communicate less that could be potentially valuable to newcomers or other test consumers.

So, since QVIs learn KERIA and SignifyTS from integration tests and since clear expectation notification behavior could help learning for newcomers I favor keeping the verbosity of the new assertions.

iFergal (approved):

.github/workflows/main.yml
kentbull: > My change in exchanging.py introduces more notifications.

Okay, single sig exn messages were not hitting Exchanger message handlers prior to #401 and thus not causing local notifications. Now, for generic outbound EXNs, #401 can introduce sender-local notifications because outbound messages now route through `agent.exc` which means their associated message handler registered in Exchanger can create notifications for those exn messages. For message handlers like IPEX this would increase local notifications because IPEX handlers do not deduplicate sender-local notifications.

Yet, multisig notifications on exn routes `/multisig/*` will have fewer notifications because `agent.exc` routes to Multiplexor and Multiplexor deduplicates sender-local notifications for multisig initiators. 

Here's the breakdown for each of the exn route types:
- `/multisig/*`: Fewer notifications due to Multiplexor deduplication
    - Event types: icp, rot, ixn, vcp, iss, rev, exn, rpy
- `/ipex/*`: More notifications since `IpexHandler` does not deduplicate and does a plain `notifier.add(...)`
    - Event types: apply, offer, agree, grant, admit, spurn
- `/challenge/response`: unchanged - `ChallengeHandler` does not add notifications. It does add a signal to Signaler.
- `/delegate/request`: Once #432 is merged, more since `DelegateRequestHandler` does a plain `notifier.add(...)`. Until #432 is merged, notification count here will be unchanged at zero.
- `/oobis`: Unchanged since `OobiRequestHandler` is not currently registered. This is likely a missing feature that works in KERIpy yet not in KERIA. It does a plain `notifier.add(...)` and would send a notification of OOBI resolution to KERIA.

kentbull: I added a more complete description of notification handling in KERIA in the keria_notifications.md document in #432: https://github.com/WebOfTrust/keria/pull/432/changes

PR #398, opened 2026-06-26
refactor: clean up equality checks and docs

kentbull:
There were a few equality checks that would possibly coerce values in an unexpected way. This PR changes double equals `==` to explicit triple equals`===` that do not do type coercion. The new tests exercise the new invalid index and ondex checks.

It also adds thorough docs to the Signer class.

### DIFF
--- a/src/keri/core/signer.ts
+++ b/src/keri/core/signer.ts
@@ -10,28 +10,68 @@ import { Siger } from './siger.ts';
 import { IdrDex } from './indexer.ts';
 import { concat } from './core.ts';
 
-/**
- * @description Signer is Matter subclass with method to create signature of serialization
- * It will use .raw as signing (private) key seed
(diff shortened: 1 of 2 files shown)


src/keri/core/signer.ts
```diff
@@ -89,23 +147,57 @@ export class Signer extends Matter {
         return this._sign(ser, this.raw, this.verfer, index, only, ondex);
     }
 
+    /**
+     * Ed25519 implementation behind `sign`.
+     *
+     * Prefer `sign()` for normal use; this method expects the caller to supply
+     * the seed and verifier that match this signing suite.
```
iFergal, line 154: imo we should make `_ed25519` private and remove this comment, better if never called externally

iFergal (commented):
Just one small comment, everything else is great!

iFergal:
also need to format this PR

PR #399, opened 2026-06-27
test: ensure IPEX op initiator is not notified

kentbull:
The initiator of an IPEX operation does not need to have a notification of an event it created. This test proves that notification is suppressed. Depends on https://github.com/WebOfTrust/keria/pull/445


kentbull:
integration tests are expected to fail until https://github.com/WebOfTrust/keria/pull/445 is merged and available for CI to target.

PR #400, opened 2026-07-02
fix: drop agent EID suffix on multisig OOBI by default

kentbull:
This has been a constant source of confusion for years and should return the intuitive result by default, the multisig Agent OOBI with no agent EID suffix.

The includeEid URL parameter allows using the old behavior, if desired.

KERIA PR: https://github.com/WebOfTrust/keria/pull/447
SignifyPy PR: https://github.com/WebOfTrust/signifypy/pull/150


kentbull:
Integration tests will fail until we get a released KERIA image we can test against. This exposes why we want to do a mono repo and merge this with the keria repo

PR #402, opened 2026-07-21
feat(parsing): add a CESR v1 stream parser

dhh1128:
## Summary

signify-ts has the CESR primitive classes (`Matter`, `Counter`, `Indexer`) and can
**emit** framed streams (`eventing.messagize`), but there is no parser that
**consumes** a stream — no equivalent of keripy's `keri.core.parsing.Parser`. This
PR adds one: `src/keri/core/parsing.ts`, a deterministic v1 CESR stream walker.

## What it does

- Frames a byte stream into messages + typed attachment groups, by the
  **version-string size** and the **attachment counters** — never by sniffing a
  leading `{`, so it works uniformly for JSON and (with an injected decoder) binary
  CBOR/MGPK bodies.
- **Delegates all sizing** to the existing `Counter` / `Matter` / `Indexer` classes
  (their `{ qb64 }` constructor + `.code`/`.count`/`.qb64`); it adds no code tables.
- Carries **byte-span provenance** (`{ start, end }`) on every node, so callers can
  map any decoded node back to the exact source bytes.
- Is **resilient**: on a code it cannot frame it stops and reports a typed error,
  returning everything parsed so far (`{ messages, errors, consumed }`) — a single
  bad byte never discards the prefix that did parse. `-V`/`-0V` material-quadlet
  wrappers are treated as resilience boundaries (their self-declared size lets the
  walk resume past undecodable inner content).
- **Pluggable body decoders**: JSON is built in; CBOR/MGPK etc. are decoded only
  when a decoder for that serialization is injected via `ParseOptions`. An undecoded
  body is still framed (`sad: null`).

Public API: `parse(bytes, opts?)`, `parseVersion(bytes, at)`, and the typed output
contract (`ParseResult`, `CesrMessage`, `AttachmentGroup`, `Primitive`,
`ParseError`, …). Wired into the barrel via one line in `src/exports.ts`.

## No new dependencies

**This PR adds zero dependencies — runtime or dev.** `package.json` and the lockfile
are untouched. `parsing.ts` imports only `./counter.ts`, `./indexer.ts`,
`./matter.ts` (modules signify-ts already ships) plus JS builtins (`TextDecoder`,
`JSON`). It reuses the primitive classes rather than adding anything.

## Excludable for consumers who don't want it

The module is **side-effect-free**: it does no work at import time (the shared
`TextDecoder` is created lazily on first use, not at module load). So a consumer
using **named imports** (`import { Matter } from 'signify-ts'`) with any
tree-shaking bundler drops `parsing.js` entirely when the parser is unused. If it
*is* included, the footprint is small — pure JS, no transitive deps.

Two caveats / optional follow-ups (maintainers' call — deliberately **not** in this PR):
- The barrel's `export default exp` (namespace) in `src/index.ts` defeats
  tree-shaking for anyone using the **default** import; named imports are unaffected.
- Adding `"sideEffects": false` to `package.json` and/or a subpath export
  (`signify-ts/parsing`) would make exclusion first-class, but those are repo-wide
  packaging decisions to validate separately.

## Testing

`test/core/parsing.test.ts` — 30 tests, inline keripy-derived vectors (matching the
existing house convention), covering: version parsing; `-A/-B/-C/-D/-E/-F/-G/-H/-I`
groups; `-V`/`-0V` nested decomposition and resilience boundaries; the Matter vs
Indexer primitive-class discriminator; and every error path (no-version-string,
malformed-body, unparseable-counter, unframable-group).

Gates: `npm run build`, `npm run lint`, `npm run pretty:check`, and
`npx vitest run test/core/parsing.test.ts` all pass. (The 4 unrelated failures in
`test/app/*` require a running KERIA agent and are pre-existing on `main`.)

## Scope

**CESR v1 only.** signify-ts's counter/version-string machinery is v1 today (no v2
count-code table, no genus/version model beyond the opaque `--AAA` counter,
`Serials` is JSON-only, `deversify`/`VEREX` are locked to the 17-char v1 version
string), and this parser matches that. The v2 additions (v2 count codes,
genus/version, a v2 version-string parser) are a separate follow-up so this PR stays
a self-contained, reviewable unit.

### DIFF
--- a/src/exports.ts
+++ b/src/exports.ts
@@ -29,6 +29,7 @@ export * from './keri/core/kering.ts';
 export * from './keri/core/manager.ts';
 export * from './keri/core/matter.ts';
 export * from './keri/core/number.ts';
+export * from './keri/core/parsing.ts';
 export * from './keri/core/prefixer.ts';
 export * from './keri/core/saider.ts';
 export * from './keri/core/salter.ts';
--- a/src/keri/core/parsing.ts
+++ b/src/keri/core/parsing.ts
@@ -1,0 +1,441 @@
+// CESR stream parser (v1): message framing + attachment-group decomposition.
+//
+// signify-ts already has the primitive classes (Matter, Counter, Indexer) and
+// emits framed streams (eventing.messagize), but has no parser that CONSUMES a
+// stream. This module fills that gap: it frames a byte stream deterministically
+// — by the version-string size and the attachment counters, never by sniffing a
+// leading `{` — delegating all primitive/counter/indexer SIZING to the existing
+// classes. It is resilient: on a code it cannot frame it stops and reports,
+// returning everything parsed so far, so a single bad byte never loses the
+// stream prefix that did parse.
+
+import { Counter } from './counter.ts';
+import { Indexer } from './indexer.ts';
+import { Matter } from './matter.ts';
+
+// Lazily-created shared decoder — constructing it at module load would be a
+// side effect that stops bundlers tree-shaking this module away for consumers
+// who never use the parser. Building it on first use keeps the module pure.
+let decoder: TextDecoder | undefined;
+const td = (): TextDecoder => (decoder ??= new TextDecoder());
+const DASH = 0x2d; // '-'
+
+/** A half-open byte range [start, end) into the original stream. */
+export interface ByteSpan {
+    start: number;
+    end: number;
+}
+
+/** Framing outcome for a node. */
+export type NodeState =
+    | 'known' // recognized code, fully framed
+    | 'unknown' // well-formed but code not recognized; may be framable by size
+    | 'invalid'; // cannot be framed
+
+/** A single CESR primitive (key, digest, signature, sequence number, …). */
+export interface Primitive {
+    kind: 'primitive';
+    code: string;
+    /** Which code table framed it — disambiguates `code`, since the Matter and
+     * Indexer tables collide (e.g. 'A' is a seed as Matter, an Ed25519 indexed
+     * signature as Indexer). */
+    class: 'matter' | 'indexer';
+    span: ByteSpan;
+}
+
+/** A counter-framed attachment group (e.g. controller sigs `-A`, quadlets `-V`). */
+export interface AttachmentGroup {
+    kind: 'group';
+    code: string;
+    count: number;
+    span: ByteSpan;
+    state: NodeState;
+    /** Typed children; empty for opaque quadlet frames and unknown codes. */
+    items: AttachmentNode[];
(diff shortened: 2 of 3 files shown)


dhh1128:
tagging @kentbull and/or @m00sey for review. Kent, this may be redundant with what you were planning; if so, I won't be offended if you just close this.

daidoji (changes requested):
I don't quit think this is correct.  Even simple pathological streams seem to fail.  Here's an example of a test that fails with a size that's bigger than the length of the stream

```
438     it('Pathological stream', () => {
439         const stream = bytesOf('{"v":"KERI10JSON100000_"}');
440         const { messages, errors } = parse(stream);
441         assert.isEmpty(messages);
442         assert.equal(errors, []);
443     });
```

dhh1128:
@daidoji : Thanks for the catch. I'm updating the code.

dhh1128:
You're right, and thanks for catching it. The parser never compares a declared size against the bytes actually available, so a truncated stream is reported as a complete one. Bad checking of preconditions on my part.

Your example reproduces exactly:

```
parse(bytesOf('{"v":"KERI10JSON100000_"}'))   // 25 bytes in
→ messages: 1, errors: [], consumed: 1048576, span: { start: 0, end: 1048576 }
```

The version string claims 0x100000 bytes. `bytes.subarray(0, 1048576)` silently clamps to the 25 that exist, that clamped slice happens to be valid JSON, and the message is emitted with a span and a `consumed` that both run a megabyte past the end of the buffer. A caller that slices by `span` gets short data with nothing in the result saying so. When the truncated body *isn't* accidentally valid JSON it reports `malformed-body`, which is also the wrong diagnosis — the bytes may be fine, there just aren't enough of them yet.

It isn't only the body. The same hole is in the attachment framing:

- a truncated `-V` wrapper is reported `state: 'known'` with a span past the end of the buffer, no error;
- a truncated `-A` group is reported `unframable-group` / `state: 'invalid'` with `permanent: true`, i.e. permanently bad, when the right answer is "come back with more bytes".

That last one is a design gap as much as a bug: `ParseError.permanent` is typed as the literal `true`, so the module currently cannot express a recoverable failure at all.

Chasing this turned up a worse one with the same root. A version string with size 0 and a serialization the parser has no decoder for makes `parse` loop forever, allocating a message per iteration:

```
parse(bytesOf('{"v":"KERI10CBOR000000_"}'))   // 25 bytes in
```

`bodyEnd === i`, no counter follows, the cursor never advances. That takes a Node process to a 4 GB heap death in about 40 seconds. In a browser it's the tab.

I've now pushed a fix.

PR #405, opened 2026-07-25
Clean up connect

kentbull:
This PR cleans up some long-standing gaps we've had in SignifyController booting and connecting:
- Adds cancellation with `AbortSignal` to boot and connect flows.
- Adds error handling to POST `/agent/<ctrl Pre>?type=ixn` delegation request.
- Reject malformed sequence data (non-string) and removes overly permissive numeric coercion.

### DIFF
--- a/src/keri/app/clienting.ts
+++ b/src/keri/app/clienting.ts
@@ -6,6 +6,7 @@ import { ExternalModule, IdentifierManagerFactory } from '../core/keeping.ts';
 import { CesrNumber } from '../core/number.ts';
 import { Tier } from '../core/salter.ts';
 import { Seqner } from '../core/seqner.ts';
+import { Serder } from '../core/serder.ts';
 
 import { Identifier } from './aiding.ts';
 import { Contacts, Challenges } from './contacting.ts';
@@ -22,6 +23,11 @@ const DEFAULT_BOOT_URL = 'http://localhost:3903';
 
 // Export type outside the class
 export type AgentResourceResult = components['schemas']['AgentResourceResult'];
+
+/** Cancellation options for booting or connecting a Signify client. */
+export interface ConnectionOptions {
+    signal?: AbortSignal;
+}
 
 class State {
     agent: any | null;
@@ -115,9 +121,10 @@ export class SignifyClient {
      * Agent already exists.
      *
      * @async
+     * @param {ConnectionOptions} [options] Cancellation options
      * @returns {Promise<Response>} Response from the KERIA boot endpoint
      */
-    async boot(): Promise<Response> {
+    async boot(options: ConnectionOptions = {}): Promise<Response> {
         const [evt, sign] = this.controller?.event ?? [];
         const data = {
             icp: evt.sad,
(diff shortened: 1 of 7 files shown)


daidoji (approved):

src/keri/app/clienting.ts
```diff
@@ -147,30 +169,141 @@ export class SignifyClient {
...
+        }
+
+        return {
+            i: this.agent.pre,
+            s: new Seqner({ sn: this.agent.sn }).snh,
+            d: this.agent.said,
+        };
+    }
+
+    /** Verify the local interaction event anchors this exact Agent. */
```
iFergal: Is this implying there is a use case to connect a client to multiple agents? If not, `sn > 0` was if anything safer as it avoids accidentally connecting to a different one.

src/keri/app/clienting.ts
```diff
@@ -147,30 +169,141 @@ export class SignifyClient {
...
+    }
+
+    /** Load agent state and initialize authenticated client services. */
+    private async connectToAgent(
+        reuseController: boolean,
+        options: ConnectionOptions
+    ) {
+        const state = await this.state(options);
+        this.pidx = state.pidx;
+        if (reuseController) {
```
iFergal: A boot is a one-time thing per client - I'm not sure what performance tests were being ran but this feels like optimizing for the wrong thing. I'm not sure it's worth the added complexity.

src/keri/app/clienting.ts
iFergal: At least until a point where we have the idea of moving to a new agent.

kentbull: No, it doesn't imply multiple agents, it just verifies the actual interaction event rather than making only a sequence number assertion.

kentbull: The main value here in my mind comes from being explicit and verifying the delegation approval. It would have helped me learn the library when I was first learning the Signify edge client and agent delegation model.

It isn't as hard to learn that these days by using AI.

I can strip this down to just the ConnectionOptions with AbortSignal if we don't believe the delegation anchor seal validation is worth the increased complexity.

src/keri/app/clienting.ts
kentbull: Avoiding re-derivation of keys in the Controller is the real target here. That activity just happens once and, on second-thought it is more of a YAGNI thing. I'm going to simplify it.

The test I was using was an end to end vLEI ecosystem test that created multiple multisig groups for the GEDA GARs, QVI QARs, and LE LARs, as well as single sig OOR and ECR holders, and used those multi and single sig AIDs in multiple operations including initial connection, OOBI resolution, challenge response, credential issuance, and IPEX operations.

This sort of test is something hardly anyone would ever run in production. Maybe partly in an integration testing or e2e testing env, yet not production.

src/keri/app/clienting.ts
iFergal: Sorry, you're right. But still, I'm unsure on the exact value of the check:
- `clienting.approveDelegation(this.agent!)` directly reads `agent.pre`, `agent.sn`, `agent.said` and creates and signs an event locally (no calls to KERIA).
- Then immediately call `_verifyAgentDelegationSeal` on the result, which compares against `agent.pre`, `agent.sn`, `agent.said`.

kentbull: Yeah, that's true. I'll remove that from this PR and later we can consider whether we want to explicitly validate, post delegation completion, the delegator and delegate KELs.

kentbull:
@lenkan will you review this?

iFergal (approved):
needs to be rebased

lenkan (approved):

PR #413, opened 2026-09-05
Feat/cesr pq codes

dhh1128:
This adds the 33 post-quantum primitive codes from CESR 1.1 to `Matter.Sizes`: FN-DSA (Falcon), ML-DSA, and SLH-DSA public keys, signatures, and private-key seeds. It's stacked on #403, so it shows those commits too until they merge. The last one is the commit to look at.

The part I'd most like feedback on is that two existing rows change meaning. In the pre-PR code, `2AAA` and `3AAA` currently hold `Sizage(4, 0, 8, 1)` and `Sizage(4, 0, 8, 2)`. They look like leftover test placeholders from an older keripy: there's no `MatterCodex` constant for either, nothing in src or test refers to them, and keripy uses `2___` and `3___` for that purpose in every version I looked at, including 1.x. CESR 1.1 gives those two codes to the FN-DSA-1024 and ML-DSA-44 public keys. I think the change is safe, but I could have missed a caller, so please let me know if either row is doing work somewhere.

I stopped at sizes and didn't add `MatterCodex` constants. keripy hasn't named these codes yet, and if signify-ts names 33 of them first, I'd expect the two to drift apart. Adding names later doesn't break anything. If you'd rather have them now, I'll put them in.

There's no change to `parsing.ts`. The parser picks these up as soon as the table has them.

The tests use two oracles, because a transcription slip in 33 rows of large numbers is otherwise invisible. One is arithmetic the spec makes checkable: each full size divides by 4, each lead size follows the selector, each raw-plus-lead falls on a 24-bit boundary, and a raw value of the implied length round-trips through `Matter` to the declared length. That last check exercises the `ls > 0` path for the 19 codes in the `2` and `3` tables.

The other oracle is real cryptography. The vector is an Ed25519 inception receipted by a witness signing with Falcon. Its FN-DSA-512 keypair and signature are genuine, made with `@noble/post-quantum` over the receipted event's exact bytes, so the signature verifies against the key sitting next to it. signify-ts has no Falcon implementation and can't check that here, but it can frame it. The generator is scripts/make-falcon-sample.mjs in bakobo/cesrview if you want to regenerate or check the vector yourself.

CESR 1.1 has no indexed post-quantum signature code, so a Falcon signature can't be a controller or witness indexed signature, and a `-C` receipt couple is the only v1 carrier for it. That's what the vector uses. There's also a test asserting that no Indexer primitive shows up in it, since a negative requirement with no failing test tends to get quietly dropped.

A couple of things I ran into look more like spec questions than PR questions. `1AAR`'s 666 bytes is the padded Falcon variant; plain falcon512 gave me 652 bytes in one run, which the code can't encode. And there's no non-transferable-prefix code for any post-quantum algorithm, only the plain verification key, so the vector puts a Falcon key where a prefix would normally go. Between that and the missing indexed code, I don't think there's a fully specified way right now to attach a post-quantum signature to a KERI event. I can raise both with the KSWG; neither affects this PR.


daidoji (approved):
PQ stuff looks good.  

2AAA and 3AAA were used a long time ago for testing cesr codes but were moved away in keripy without fanfare.  https://github.com/WebOfTrust/keripy/blame/6473d7efe811254763986555dc1d9873b9d8524d/src/keri/core/coring.py#L318

